PR #1, opened 2016-08-11
travis ci

cheecheeo:
Basic travis ci script.


seantakeda:
@cheecheeo, thanks for creating one. Actually the supported Pythons are 2.7.9+/3.4.3+, so other versions should not be validated. It would be nice to remove others.

cheecheeo:
@seantakeda please take another look.

PR #2, opened 2016-08-23
Badges FTW!

MarcinZukowski:


cheecheeo:
@MarcinZukowski looks good to me.

PR #4, opened 2016-10-06
Fix tests

smtakeda:
Attempting to fix travis tests. So far no concurrent test can run as the hardcoded DB and schema is used. Will update that later.


smtakeda:
It appeared PR across separate repositories didn't work.

PR #5, opened 2016-10-06
Fix travis tests

smtakeda:
Second try to fix travisCI tests.


smtakeda:
Still the tests cannot run in parallel. Will fix in the next round of commit.

PR #8, opened 2016-10-07
Enabled verbose mode in pytest

smtakeda:


smtakeda:
There is some problem in concurrent test runs. Limited to up to 2 at the moment.

PR #13, opened 2017-04-10
map `DATETIME` to `TIMESTAMP(0)`

nbrachet:
`DATETIME` isn't supported by snowflake:
```
create table nick (
  test DATETIME
)
```
results in
```
SQL compilation error: Unsupported data type 'TOK_DATETIME'.
```

So map `DATETIME` to `TIMESTAMP_TZ(0)` or `TIMESTAMP_NTZ(0)` (whether the `DateTime` column specifies `timestamp=True` or `timestam=False`)

### DIFF
--- a/base.py
+++ b/base.py
@@ -225,6 +225,13 @@ class SnowflakeTypeCompiler(compiler.GenericTypeCompiler):
     def visit_OBJECT(selfself, type, **kw):
         return "OBJECT"
 
+    def visit_DATETIME(selfself, type, **kw):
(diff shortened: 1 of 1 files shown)


smtakeda:
Thanks for the patch. Looking now.

smtakeda:
Snowflake 1.89 will have `DATETIME` support as alias of `TIMESTAMP_NTZ`, so once the new server version is deployed, this should not be an issue.

NOTE the deployment is scheduled this afternoon.

PR #14, opened 2017-04-10
add support for `CREATE TABLE … CLUSTER BY`

nbrachet:
Add support for adding `CLUSTER BY` syntax to `CREATE TABLE`:
```
        >>> import sqlalchemy as sa
        >>> from sqlalchemy.schema import CreateTable
        >>> engine = sa.create_engine('snowflake://om1')
        >>> metadata = sa.MetaData()
        >>> user = sa.Table(
        ...     'user',
        ...     metadata,
        ...     sa.Column('id', sa.Integer, primary_key=True),
        ...     sa.Column('name', sa.String),
        ...     snowflake_clusterby=['id', 'name']
        ... )
        >>> print(CreateTable(user).compile(engine))

        CREATE TABLE "user" (
            id INTEGER NOT NULL AUTOINCREMENT,
            name VARCHAR,
            PRIMARY KEY (id)
        ) CLUSTER BY (id, name)


```

### DIFF
--- a/base.py
+++ b/base.py
@@ -10,6 +10,7 @@ import sqlalchemy.types as sqltypes
 from sqlalchemy import exc as sa_exc
 from sqlalchemy import util as sa_util
 from sqlalchemy.engine import default, reflection
+from sqlalchemy.schema import Table
 from sqlalchemy.sql import (
     compiler, expression)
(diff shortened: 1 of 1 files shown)


smtakeda:
Looking into the patch too...

PR #16, opened 2017-05-15
Adding COPY reserved word into base.py

Pangstar:
Without this snowflake users cannot execute a copy statement from sql-alchemy.


smtakeda:
Verified it is reproducible. Adding test cases.

PR #20, opened 2017-12-19
Add get_schema_names

andrewsali:
This PR adds an implementation for the [`get_schema_names`](http://docs.sqlalchemy.org/en/latest/core/reflection.html#sqlalchemy.engine.reflection.Inspector.get_schema_names) method.

Without this implementation, `get_schema_names` returns an empty list. This causes issues in derived applications such as [Superset](https://github.com/apache/incubator-superset).

### DIFF
--- a/base.py
+++ b/base.py
@@ -654,6 +654,13 @@ SELECT /* sqlalchemy:get_columns */
 
         return ret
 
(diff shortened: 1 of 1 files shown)


smtakeda:
Looking...

Issue #30, opened 2018-01-02
DESCRIBE TABLE result is not cached for get_columns and get_primary_keys


cladden:
As far was I can tell, the exact same `DESCRIBE TABLE ...` command is called in both `get_columns()` and `get_primary_keys()`. During reflection, this causes negative performance issues. If DESCRIBE TABLE has to be called twice, could the result be cached to reduce one of the calls?

As an example, if a new caching method stored the `DESCRIBE TABLE ...` n2i, result:

```
@reflection.cache
def describe_table(self, connection, full_table_name, **kw):
    result = connection.execute(
            "DESCRIBE TABLE /* sqlalchemy */ {0}".format(
                    full_table_name))
    n2i = self.__class__._map_name_to_idx(result)
    result = [row for row in result]
    return n2i, result
```

Then, it could potentially be used in this manner...

In `get_primary_keys()`:
```
@reflection.cache
def get_primary_keys(self, connection, table_name, schema=None, **kw):
    schema = schema or self.default_schema_name
    if not schema:
        row = connection.execute("SELECT CURRENT_SCHEMA()").fetchone()
        schema = self.normalize_name(row[0])
    full_table_name = self._denormalize_quote_join(schema, table_name)

    n2i, result = self.describe_table(connection, full_table_name, **kw)
    ...
```

In `get_columns()`:
```
@reflection.cache
def get_columns(self, connection, table_name, schema=None, **kw):
    schema = schema or self.default_schema_name
    if not schema:
        row = connection.execute("SELECT CURRENT_SCHEMA()").fetchone()
        schema = self.normalize_name(row[0])
    full_table_name = self._denormalize_quote_join(schema, table_name)

    n2i, result = self.describe_table(connection, full_table_name, **kw)
    ...
```

I'm sure there's probably cleaner ways than the above to implement the caching and/or refactor to reduce the repeated code blocks :-)

smtakeda:
can you take a look at the latest update?
https://github.com/snowflakedb/snowflake-sqlalchemy/commit/afb589537c9bd176de0334e6790ff7d2ad7021b0

Regression Tests passed. will push it part of v1.0.9 anyway.

cladden:
I left a few comments on https://github.com/snowflakedb/snowflake-sqlalchemy/commit/afb589537c9bd176de0334e6790ff7d2ad7021b0 

[...]

Issue #29, opened 2018-01-02
get_foreign_keys should cache database and schema


cladden:
As noted in the `SnowflakeDialect` comments, the current implementation of `get_foreign_keys()` does not cache the database or schema. Instead, it goes out to snowflake to obtain the `full_schema_name ` for every table, resulting in 2 calls to snowflake per table. 

This is problematic for alembic reflection of schemas with many tables. snowflake reflection can take as long as 15 minutes; while redshift/postgres is under 1 minute. Any caching that can reduce excess calls to snowflake is greatly appreciated :-). 

For reference, using cProfile during reflection shows:
* 66.9% of the time is spent in `get_columns()`
* 18.5% of the time is spent in `get_foreign_keys()`
* 6.7% of the time is spent in `get_primary_keys()`

The low hanging fruit may be to address the 18.5% issue if the `SnowflakeDialect` introduced a new caching method:

```
@reflection.cache
def get_full_schema_name(self, connection, schema=None, **kw):
    schema = schema or self.default_schema_name
    row = connection.execute("SELECT CURRENT_DATABASE(), "
                             "CURRENT_SCHEMA()").fetchone()
    full_schema_name = self._denormalize_quote_join(
            row[0], schema if schema else row[1])
    return full_schema_name
```

Then it could be used like this in `get_foreign_keys()`:
```
@reflection.cache
def get_foreign_keys(self, connection, table_name, schema=None, **kw):
    """
    Gets all foreign keys
    """

    full_schema_name = self.get_full_schema_name(connection, schema, **kw)
    ...
```

smtakeda:
Can you please provide Snowflake implementation in`env.py` and `alembic` command that causes the issue?

I had played with `alembic` a while back, but honestly I'm not familiar with the tool, thought I know the concept of DB schema management. It would be great if I know your use case to start looking into.

cladden:
alembic command that migrated the database:
```
ALEMBIC_DIR="/srv/app/database"
PYTHONPATH=${ALEMBIC_DIR} alembic -c ${ALEMBIC_DIR}/alembic.ini
```

env.py:
```
from __future__ import with_statement

import os
import sys
from logging.config import fileConfig

from alembic import context
from alembic_support.db_env import make_url_from_env
from alembic_support.db_env import get_engine_from_env_config
from alembic_support.configure_context import configure_offline
from alembic_support.configure_context import configure_online

# this is the Alembic Config object, which provides
[...]

PR #33, opened 2018-01-03
Fixed reflection cache

smtakeda:
The method should have included kw parameters.

### DIFF
--- a/base.py
+++ b/base.py
@@ -423,15 +423,14 @@ class SnowflakeDialect(default.DefaultDialect):
         return '.'.join(
             self.identifier_preparer._quote_free_identifiers(*idents))
 
-    @reflection.cache
-    def _current_database_schema(self, connection, **_):
-        row = connection.execute(
-            "SELECT /* sqlalchemy:_current_database_schema */ "
-            "CURRENT_DATABASE(), CURRENT_SCHEMA()").fetchone()
-        return self.normalize_name(row[0]), self.normalize_name(row[1])
+    def _current_database_schema(self, connection):
+        return (
+            self.normalize_name(connection.connect().connection.database),
+            self.normalize_name(connection.connect().connection.schema))
 
-    def _get_default_schema_name(self, connection, **kw):
-        _, current_schema = self._current_database_schema(connection, **kw)
+    def _get_default_schema_name(self, connection):
+        # NOTE: no cache object is passed here
+        _, current_schema = self._current_database_schema(connection)
         return current_schema
 
     @staticmethod
@@ -461,7 +460,7 @@ class SnowflakeDialect(default.DefaultDialect):
     def get_primary_keys(self, connection, table_name, schema=None, **kw):
         schema = schema or self.default_schema_name
         if not schema:
-            _, schema = self._current_database_schema(connection, **kw)
+            _, schema = self._current_database_schema(connection)
 
         full_table_name = self._denormalize_quote_join(schema, table_name)
 
--- a/base.py
+++ b/base.py
@@ -424,9 +424,10 @@ class SnowflakeDialect(default.DefaultDialect):
             self.identifier_preparer._quote_free_identifiers(*idents))
 
     def _current_database_schema(self, connection):
+        con = connection.connect().connection
         return (
-            self.normalize_name(connection.connect().connection.database),
-            self.normalize_name(connection.connect().connection.schema))
+            self.normalize_name(con.database),
+            self.normalize_name(con.schema))
 
     def _get_default_schema_name(self, connection):
         # NOTE: no cache object is passed here
--- a/base.py
+++ b/base.py
@@ -5,6 +5,7 @@
 #
 
 import re
+from collections import OrderedDict
 
 import sqlalchemy.types as sqltypes
 from sqlalchemy import exc as sa_exc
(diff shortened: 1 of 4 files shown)


base.py
```diff
@@ -589,16 +577,104 @@ def get_columns(self, connection, table_name, schema=None, **kw):
...
 
-            columns.append(cdict)
+    @reflection.cache
+    def get_columns(self, connection, table_name, schema=None, **kw):
+        """
+        Gets all column info given the table info
+        """
+        schema = schema or self.default_schema_name
+        if not schema:
+            _, schema = self._current_database_schema(connection)
```
cladden: is this missing the **kw argument for info_cache?

cladden: same question about _current_database_schema in get_foreign_keys

base.py
```diff
@@ -607,15 +683,22 @@ def get_table_names(self, connection, schema=None, **kw):
...
         schema = schema or self.default_schema_name
+        current_schema = schema
         if schema:
             cursor = connection.execute(
                 "SHOW /* sqlalchemy:get_table_names */ TABLES IN {0}".format(
                     self._denormalize_quote_join(schema)))
         else:
             cursor = connection.execute(
                 "SHOW /* sqlalchemy:get_table_names */ TABLES")
+            _, current_schema = self._current_database_schema(connection)
```
cladden: should this call include the **kw argument to provide the info_cache?

cladden:
I was able to profile this PR branch and I'm seeing the caching not working for `_current_database_schema()` and it's not clear if it's working as intended for `get_foreign_keys()` / `get_columns()`. I left a few comments for possible places to investigate.

Summary:
* `get_columns()` calls `execute()` once per table and is by far the most time consuming function
* `_current_database_schema()` calls execute() once per table 
* `get_foreign_keys()` _may_ be calling `execute()` once per table

myprofile

base.py
```diff
@@ -367,6 +368,8 @@ def create_connect_args(self, url):
             opts['port'] = '443'
         opts['autocommit'] = False  # autocommit is disabled by default
         opts.update(url.query)
+        self._cache_column_metadata = opts.get('cache_column_metadata',
```
cladden: nit: If `cache_column_metadata` is not a standard sqlalchemy option, it may be helpful to others if the option was documented somewhere.

base.py
smtakeda: my mistake too. Fixing.

base.py
smtakeda: Yes, if the cache works well, I'll update the doc too.

base.py
smtakeda: Carefully looking into `_current_database_schema`, it appeared this won't work as the key is not given.

base.py
```diff
@@ -423,15 +423,14 @@ def _denormalize_quote_join(self, *idents):
...
-            "SELECT /* sqlalchemy:_current_database_schema */ "
-            "CURRENT_DATABASE(), CURRENT_SCHEMA()").fetchone()
-        return self.normalize_name(row[0]), self.normalize_name(row[1])
-
-    def _get_default_schema_name(self, connection, **kw):
-        _, current_schema = self._current_database_schema(connection, **kw)
+    def _current_database_schema(self, connection):
+        return (
+            self.normalize_name(connection.connect().connection.database),
+            self.normalize_name(connection.connect().connection.schema))
```
smtakeda: Use cached current database and schema.

Those attributes are updated on every query, so they should be up to date.

base.py
```diff
@@ -461,7 +460,7 @@ def _describe_table(self, connection, full_table_name, **_):
     def get_primary_keys(self, connection, table_name, schema=None, **kw):
         schema = schema or self.default_schema_name
         if not schema:
-            _, schema = self._current_database_schema(connection, **kw)
+            _, schema = self._current_database_schema(connection)
```
smtakeda: no inspector cache is used.

base.py
```diff
@@ -423,15 +423,14 @@ def _denormalize_quote_join(self, *idents):
...
-    def _get_default_schema_name(self, connection, **kw):
-        _, current_schema = self._current_database_schema(connection, **kw)
+    def _current_database_schema(self, connection):
+        return (
+            self.normalize_name(connection.connect().connection.database),
+            self.normalize_name(connection.connect().connection.schema))
+
+    def _get_default_schema_name(self, connection):
+        # NOTE: no cache object is passed here
+        _, current_schema = self._current_database_schema(connection)
```
smtakeda: This method is never called with **kw.

base.py
```diff
@@ -589,16 +576,104 @@ def get_columns(self, connection, table_name, schema=None, **kw):
...
                 'type': type_instance,
                 'nullable': is_nullable == 'YES',
                 'default': column_default,
                 'autoincrement': is_identity == 'YES',
-                'primary_key': column_map[colname],
             }
 
-            columns.append(cdict)
+    @reflection.cache
+    def get_columns(self, connection, table_name, schema=None, **kw):
```
cladden: is there's any way to avoid/reduce the code duplication with _get_columns_for_tables? This is a complicated method and it may be easier to maintain if there was a single copy.

smtakeda: sure, will do the refactoring later today.

cladden:
I just ran the latest in the PR and the caching helps reduce reflection time from 7 minutes to ~2.5 minutes. Awesome!

smtakeda:
Fixed #30 and #29

PR #36, opened 2018-01-23
added role paramer to doc

### LINKED ISSUE #35: Connection URL parameters aren't documented anywhere
I'm developing with snowflake-sqlalchemy, and in my organization control over the snowflake `ROLE` is important. I couldn't find any documentation either on snowflake's website, the snowflake-connector-python docs, or the snowflake-sqlalchemy docs indicating that it might be possible to specify role in a connection URL. ~~However, it looks like adding `?roleName=` to the connection URL would do the trick, based on the [python connector code](https://github.com/snowflakedb/snowflake-connector-python/blob/6c3e666ef6c388a3bde2b190256f88a615259196/auth.py#L156).~~ As mentioned below, this was incorrect, and `role` is the proper parameter. 

Where should this be documented? I feel as if this repo is the appropriate place, as it contains the url constructor util and is also a heavy user of the URLs. Would it be appropriate to submit a PR documenting the ~~`roleName`~~ (and possibly other) url connection parameters?

smtakeda:
Fixed #35

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -98,23 +98,23 @@ Where:
 - :code:`<password>` is the password for your Snowflake user.
 - :code:`<account_name>` is the name of your Snowflake account.
 
-You can optionally specify the initial database and schema for the Snowflake session by including them at the end of the connection string, separated by :code:`/`. You can also specify the initial warehouse for the session as a parameter string at the end of the connection string:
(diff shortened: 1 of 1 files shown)


kheuton:
Looks great, I've tested it and it fixed my problem. Thanks for your help!

Issue #38, opened 2018-01-29
Invalid return value for SnowflakeDialect.get_primary_keys (deprecated)


nrth:
Python 2.7.12, Linux-4.4.0-112-generic-x86_64-with-Ubuntu-16.04-xenial
> SQLAlchemy==1.1.15
> snowflake-sqlalchemy==1.0.9
> snowflake-connector-python==1.4.17

I tracked down an issue with incorrectly set primary keys on reflection lookups.

#### Steps to reproduce
1. Setup a table/schema with primary key column
2. Do a lookup with sqlalchemy.Table.columns
3. The primary_key property of the pk column is incorrect

#### Possible solution

`base.SnowflakeDialect.get_primary_keys` implements a [deprecated](http://docs.sqlalchemy.org/en/latest/core/reflection.html#sqlalchemy.engine.reflection.Inspector.get_primary_keys) method and should return a list of primary keys if implemented (returns a constrained_columns dict currently).

`get_pk_constraint` is the [replacement](http://docs.sqlalchemy.org/en/latest/core/reflection.html#sqlalchemy.engine.reflection.Inspector.get_pk_constraint) for `get_primary_keys`, the compat implementation in the super class is still based on `get_primary_keys`, the invalid return value of `get_primary_keys` results in an incorrect dict though:

> {'constrained_columns': {'constrained_columns': ['id'], 'name': None}}

Expected:

> {'constrained_columns': ['id'], 'name': None}

So basically renaming `get_primary_keys` to `get_pk_constraint` should do the trick, happy to submit a pull request if needed.

smtakeda:
Thanks for catching the issue. I knew `get_primary_keys` was deprecated and was thinking of replacing with `get_pk_constraint` soon but didn't realize the return values are different...
The fix looks good, but the existing applications that depend on a wrong behavior of `get_primary_keys` will be impacted and have to replace with `get_pk_constraint`.

Anyway I know how to fix it and will update the test cases as well.

PR #39, opened 2018-01-29
Fixed the return value of get_pk_constraint. 

smtakeda:
Note this fix will change the return value of get_primary_keys. The existing applications that depend on a wrong behavior of get_primary_keys should replace with get_pk_constraint


smtakeda:
Fixing #38

PR #48, opened 2018-05-09
Fix #47

### LINKED ISSUE #47: Best practice for doing cross database joins using ORM
### Question

Hi there, in snowflake we often have multiple databases, with their own set of schemas, and tables inside those schemas. Ideally data shouldn't have to move between these databases nor schemas, but in practice they often do.

One use case is when there might be valuable data between two different databases. In my example, I'm going to reference database `A` and database `B` which has schemas `ant` and `babboons`, which has tables `apples` and `bananas` respectively.

Through a direct query using snowflake SQL, you could implement a join like
```
select
    *
from A.ant.apples a
join B.babboons.bananas b
    on a.id = b.id
```

How can you implement a cross database join like the above example using a single session through the ORM interface? my understanding is that a session can only point towards one database.

The query would look like such:
```
# assume session was created with database=A
query = session.query(Ant.id, Banana.id)\
    .join(Banana, Ant.id == Banana.id)
```

This would obviously fail because the query would search for a A.babboons.banana (Database A).

Can this be done through the ORM interface without dropping to SQL? If so, what's the best practice in doing this?

### Issue Template

1. What version of Python are you using (`python --version`)?

Python 2.7.12

2. What operating system and processor architecture are you using (`python -c 'import platform; print(platform.platform())'`)?

Linux-4.9.76-3.78.amzn1.x86_64-x86_64-with-glibc2.2.5

3. What are the component versions in the environment (`pip list`)? 

snowflake-connector-python==1.5.8
snowflake-sqlalchemy==1.1.0

4. What did you do?
If possible, provide a recipe for reproducing the error.
A complete runnable program is good.

N/A

6. What did you see instead?

N/A

7. Can you set logging to DEBUG and collect the logs?

N/A

smtakeda:
Split schema name into db and schema if dot separated, and add quotes to each if required, then reconstruct a full namespace.

Fixed #47

### DIFF
--- a/base.py
+++ b/base.py
@@ -164,8 +164,41 @@ class SnowflakeIdentifierPreparer(compiler.IdentifierPreparer):
         """
         Split schema by a dot and merge with required quotes
         """
-        idents = schema.split('.')
+        idents = self._split_schema_by_dot(schema)
         return '.'.join(self._quote_free_identifiers(*idents))
+
+    def format_label(self, label, name=None):
+        """
+        Format label after removing double quotes, which are not currently
+        supported in a object identifier.
+        """
+        s = (name or label.name).replace(self.escape_quote, '')
+        return self.quote(s)
+
+    def _split_schema_by_dot(self, schema):
+        ret = []
+        idx = 0
+        pre_idx = 0
+        in_quote = False
+        while idx < len(schema):
+            if not in_quote:
+                if schema[idx] == '.' and pre_idx < idx:
+                    ret.append(schema[pre_idx:idx])
+                    pre_idx = idx + 1
+                elif schema[idx] == '"':
+                    in_quote = True
+                    pre_idx = idx + 1
+            else:
+                if schema[idx] == '"' and pre_idx < idx:
+                    ret.append(schema[pre_idx:idx])
+                    in_quote = False
+                    pre_idx = idx + 1
+            idx += 1
+            if pre_idx < len(schema) and schema[pre_idx] == '.':
+                pre_idx += 1
+        if pre_idx < idx:
+            ret.append(schema[pre_idx:idx])
+        return ret
 
 
 class SnowflakeCompiler(compiler.SQLCompiler):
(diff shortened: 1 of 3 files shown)


base.py
```diff
@@ -158,6 +160,13 @@ def _quote_free_identifiers(self, *ids):
         """
         return tuple([self.quote(i) for i in ids if i is not None])
 
+    def quote_schema(self, schema, force=None):
+        """
+        Split schema by a dot and merge with required quotes
+        """
+        idents = schema.split('.')
+        return '.'.join(self._quote_free_identifiers(*idents))
```
smtakeda: By default, this method takes schema as one identifier. The fix here is split and merge with required quotes.

base.py
```diff
@@ -420,8 +428,12 @@ def denormalize_name(self, name):
         return name
 
     def _denormalize_quote_join(self, *idents):
+        split_idents = reduce(
+            operator.add,
+            [ids.split('.') for ids in idents if ids is not None])
+        split_idents = [ids.strip('"') for ids in split_idents]
```
smtakeda: A element of the list of identifiers may be already reconstructed, so split them again here.

kevinjfoley:
I took a quick look and it looks good for our use case, also installed locally and confirmed it's working.

One potential issue is I don't think it properly handles the case where there is a `.` in the database or schema name.  We don't have anything like that but per the documentation it seems like `.`'s are allowed as long as the name is quoted.

For example I tested:

```
class User(Base):
    __tablename__ = 'users'
    __table_args__ = {'schema': '"dot.test"'}
    id = Column(Integer, primary_key=True)

query = session.query(User.id)
print(str(query))
``` 

which resulted in 

```
SELECT """dot"."test""".users.id AS """dot_test""_users_id" 
FROM """dot"."test""".users
```

kevinjfoley:
We don't have anything like it and I've never seen it anywhere so I think it's safe to leave out.

My main concern was that it would have worked previously and this change broke things but I tested against the current release of the package and it isn't working currently so I don't think that's a problem.

smtakeda:
The latest change includes:
- Support schema, db names including dot characters
- Remove double quotes from label as Snowflake doesn't support it in object identifiers.

Issue #57, opened 2018-08-25
Regression in handling of quoted_name in compiler after 1.1.0


villebro:
When passing a select query to the compiler, quotes around the alias are lost. This was working correctly in v. 1.1.0, but stopped working as of 1.1.1

Please answer these questions before submitting your issue. Thanks!

1. What version of Python are you using (`python --version`)?
3.6.4

2. What operating system and processor architecture are you using (`python -c 'import platform; print(platform.platform())'`)?
Darwin-17.7.0-x86_64-i386-64bit

3. What are the component versions in the environment (`pip list`)? 
```
Package                    Version  
-------------------------- ---------
asn1crypto                 0.24.0   
azure-common               1.1.14   
azure-nspkg                2.0.0    
azure-storage              0.36.0   
boto3                      1.7.84   
botocore                   1.10.84  
certifi                    2018.8.24
cffi                       1.11.5   
chardet                    3.0.4    
cryptography               2.3.1    
docutils                   0.14     
future                     0.16.0   
idna                       2.7      
ijson                      2.3      
jmespath                   0.9.3    
pip                        18.0     
pyasn1                     0.4.4    
pyasn1-modules             0.2.2    
pycparser                  2.18     
pycryptodomex              3.6.6    
PyJWT                      1.6.4    
pyOpenSSL                  18.0.0   
python-dateutil            2.7.3    
pytz                       2018.5   
requests                   2.19.1   
s3transfer                 0.1.13   
setuptools                 40.2.0   
six                        1.11.0   
snowflake-connector-python 1.6.7    
snowflake-sqlalchemy       1.1.2    
SQLAlchemy                 1.2.11   
urllib3                    1.23     
wheel                      0.31.1
```
4. What did you do?
If possible, provide a recipe for reproducing the error.
A complete runnable program is good.
```
import sqlalchemy as sa
from sqlalchemy.sql import quoted_name, label, column, table

engine = sa.create_engine(‘snowflake://XXX’)

col = column('colname').label(quoted_name('alias', True))
select_from_table = sa.select([col]).group_by(col).select_from(table(quoted_name('abc', True)))
compiled_result = sel_from_tbl.compile(engine)

print(compiled_result)
```

5. What did you expect to see?
```
SELECT colname AS "alias" 
FROM "abc" GROUP BY colname
```

6. What did you see instead?
```
SELECT colname AS alias 
FROM "abc" GROUP BY colname
```

[...]

PR #67, opened 2018-11-08
preserve quote label

smtakeda:
Fixing #57.

### DIFF
--- a/test/test_compiler.py
+++ b/test/test_compiler.py
@@ -74,6 +74,15 @@ FROM "abc" GROUP BY colname"""
     assert str(compiled_result) == """SELECT colname AS alias 
 FROM "abc" GROUP BY colname"""
 
+    # Not quote label in capital case
+    col = column('colname').label('ALIAS')
+    sel_from_tbl = select([col]).group_by(col).select_from(
+        table(quoted_name('abc', True)))
+    compiled_result = sel_from_tbl.compile(engine_testaccount)
+
+    assert str(compiled_result) == """SELECT colname AS "ALIAS" 
+FROM "abc" GROUP BY colname"""
+
     # Not quote table
     col = column('colname').label('alias')
     sel_from_tbl = select([col]).group_by(col).select_from(
--- a/test/test_compiler.py
+++ b/test/test_compiler.py
@@ -56,38 +56,20 @@ class TestSnowflakeCompiler(AssertsCompiledSQL):
 
 
 def test_quoted_name_label(engine_testaccount):
-    # quote label
-    col = column('colname').label(quoted_name('alias', True))
-    sel_from_tbl = select([col]).group_by(col).select_from(
-        table(quoted_name('abc', True)))
-    compiled_result = sel_from_tbl.compile(engine_testaccount)
+    test_cases = [
+        # quote name
+        {"label": quoted_name('alias', True),
+         "output": 'SELECT colname AS "alias" \nFROM abc GROUP BY colname'},
+        # not quote label
+        {"label": 'alias',
+         "output": 'SELECT colname AS alias \nFROM abc GROUP BY colname'},
+        # not quote mixed case label
+        {"label": 'Alias',
+         "output": 'SELECT colname AS "Alias" \nFROM abc GROUP BY colname'},
+    ]
 
-    assert str(compiled_result) == """SELECT colname AS "alias" 
-FROM "abc" GROUP BY colname"""
-
-    # Not quote label
-    col = column('colname').label('alias')
-    sel_from_tbl = select([col]).group_by(col).select_from(
-        table(quoted_name('abc', True)))
-    compiled_result = sel_from_tbl.compile(engine_testaccount)
-
-    assert str(compiled_result) == """SELECT colname AS alias 
-FROM "abc" GROUP BY colname"""
-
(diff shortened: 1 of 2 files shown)


smtakeda:
@villebro, would you please review this change?

villebro:
@smtakeda sure, would be happy to do so, will review today.

test/test_compiler.py
```diff
@@ -46,11 +45,49 @@ def test_multi_table_delete_multiple(self):
...
+FROM "abc" GROUP BY colname"""
+
+    # Not quote label
+    col = column('colname').label('alias')
+    sel_from_tbl = select([col]).group_by(col).select_from(
+        table(quoted_name('abc', True)))
+    compiled_result = sel_from_tbl.compile(engine_testaccount)
+
+    assert str(compiled_result) == """SELECT colname AS alias 
+FROM "abc" GROUP BY colname"""
```
villebro: For the sake of clarity, I would not quote the table name in this case:

```python
    # Not quote label
    col = column('colname').label('alias')
    sel_from_tbl = select([col]).group_by(col).select_from('abc')
    compiled_result = sel_from_tbl.compile(engine_testaccount)
     assert str(compiled_result) == """SELECT colname AS alias 
FROM abc GROUP BY colname"""
```

test/test_compiler.py
```diff
@@ -46,11 +45,49 @@ def test_multi_table_delete_multiple(self):
...
+
+    # Not quote label in capital case
+    col = column('colname').label('ALIAS')
+    sel_from_tbl = select([col]).group_by(col).select_from(
+        table(quoted_name('abc', True)))
+    compiled_result = sel_from_tbl.compile(engine_testaccount)
+
+    assert str(compiled_result) == """SELECT colname AS "ALIAS" 
+FROM "abc" GROUP BY colname"""
+
```
villebro: I believe having a mixed-case test case might be better, as all uppercase in this case would in fact not impact the results (do correct me if I am wrong). So perhaps the following might be more important to test. Also here remove quoting from table name:

```python
    # Not quote label in mixed case
    col = column('colname').label('Alias')
    sel_from_tbl = select([col]).group_by(col).select_from('abc')
    compiled_result = sel_from_tbl.compile(engine_testaccount)
     assert str(compiled_result) == """SELECT colname AS "Alias" 
FROM abc GROUP BY colname"""
```

villebro:
@smtakeda I added a few comments, but overall LGTM.

test/test_compiler.py
```diff
@@ -46,11 +45,49 @@ def test_multi_table_delete_multiple(self):
...
+def test_quoted_name_label(engine_testaccount):
+    # quote label
+    col = column('colname').label(quoted_name('alias', True))
+    sel_from_tbl = select([col]).group_by(col).select_from(
+        table(quoted_name('abc', True)))
+    compiled_result = sel_from_tbl.compile(engine_testaccount)
+
+    assert str(compiled_result) == """SELECT colname AS "alias" 
+FROM "abc" GROUP BY colname"""
+
```
villebro: For the sake of clarity, I would not quote the table name in this case:

```python
    # quote label
    col = column('colname').label(quoted_name('alias', True))
    sel_from_tbl = select([col]).group_by(col).select_from(
        table('abc'))
    compiled_result = sel_from_tbl.compile(engine_testaccount)
     assert str(compiled_result) == """SELECT colname AS "alias" 
FROM abc GROUP BY colname"""
```

test/test_compiler.py
```diff
@@ -46,11 +45,49 @@ def test_multi_table_delete_multiple(self):
...
+    # Not quote table
+    col = column('colname').label('alias')
+    sel_from_tbl = select([col]).group_by(col).select_from(
+        table('abc'))
+    compiled_result = sel_from_tbl.compile(engine_testaccount)
+
+    assert str(compiled_result) == """SELECT colname AS alias 
+FROM abc GROUP BY colname"""
```
villebro: This test case I would remove, unless there is an explicit need to test quoting of table names.

test/test_compiler.py
smtakeda: ok, but your test case in #57 did the quote the table name?

test/test_compiler.py
smtakeda: makes sense.

test/test_compiler.py
villebro: Yeah, I don't remember exactly what my reasoning was, but the test case was written in just a few minutes and I think my point was to show that it was quoting the table name, but not the label. But to keep things as modular as possible, I would keep the label quoting test cases separate from any potential table name quoting tests.

test/test_compiler.py
smtakeda: ok, sounds good.

test/test_compiler.py
smtakeda: Fair enough.

smtakeda:
@villebro, thanks for reviewing the code!

villebro:
Thanks @smtakeda, I like the latest changes!

tomsej:
Hi, thanks for this! Do you know when v1.1.4 will be released?

smtakeda:
v1.1.4 will be released next Tuesday. Thanks.

PR #71, opened 2018-12-10
supports_multivalues_insert should be true

danfrankj:
Documentation: https://docs.snowflake.net/manuals/sql-reference/sql/insert.html 
and manual testing suggest that snowflake supports multivalue inserts

### DIFF
--- a/base.py
+++ b/base.py
@@ -423,6 +423,9 @@ class SnowflakeDialect(default.DefaultDialect):
 
     # The dialect supports a native ENUM construct.
     supports_native_enum = False
+    
+    # The dialect supports inserting multiple rows at once 
+    supports_multivalues_insert = True
 
     preparer = SnowflakeIdentifierPreparer
     ddl_compiler = SnowflakeDDLCompiler


danfrankj:
cc: @kandouss

smtakeda:
@danfrankj I can add the flag, but in my experiment there was not effect.
See #55. Did you see any improvement by adding it?

danfrankj:
@smtakeda I'm using sqlalchemy to perform an insert via 
```
conn.execute(table.insert(data))  # multivalues insert
```
and sqlalchemy errors because this flag is not set 

```
/usr/local/lib/python3.6/site-packages/sqlalchemy/sql/compiler.py in visit_insert(self, insert_stmt, asfrom, **kw)
   2073                     "version settings does not support "
   2074                     "in-place multirow inserts." %
-> 2075                     self.dialect.name)
   2076             crud_params_single = crud_params[0]
   2077         else:

CompileError: The 'snowflake' dialect with current database version settings does not support in-place multirow inserts.
```

With single inserts `conn.execute(table.insert(), data)` , I still am seeing snowflake execute multi-inserts, so the performance should be the same.

PR #94, opened 2019-04-22
SNOW-73545: Test cases for numpy datetime64 with timezone

smtakeda:


smtakeda:
@ankit-bhatnagar167 , we can add this to master but won't need to release SQLAlchemy dialect.

smtakeda:
merging this to prod branch.

PR #95, opened 2019-05-01
Move azure blobs to extras_require from setup.py

richiverse:
I'm not using Azure in my environment so this should allow users to install the necessary libraries if they do require it.

---
This change is [Reviewable](https://reviewable.io/reviews/snowflakedb/snowflake-sqlalchemy/95)


richiverse:
Closing and moving to snowflake-connector. Apologies

Issue #99, opened 2019-05-13
CopyInto's maxfilesize method expects a bool instead of an int


colemandsmith:
Please answer these questions before submitting your issue. Thanks!

1. What version of Python are you using (`python --version`)?

Python 3.6.8 :: Anaconda, Inc.

2. What operating system and processor architecture are you using (`python -c 'import platform; print(platform.platform())'`)?

Darwin-18.5.0-x86_64-i386-64bit

3. What are the component versions in the environment (`pip list`)? 

snowflake-connector-python 1.7.11
snowflake-sqlalchemy 1.1.12

4. What did you do?
If possible, provide a recipe for reproducing the error.
A complete runnable program is good.

For a `CopyInto` object I made, I tried to set the maxfilesize to 50 MB (int value of 50000000) like so:

```
unload_stmt = CopyInto(table, bucket, formatter) 
unload_stmt.maxfilesize(50000000)
```

This was based on my understanding of the [snowflake docs](https://docs.snowflake.net/manuals/sql-reference/sql/copy-into-location.html#label-copy-into-location-copyoptions) that CSVs could be set to have a max file size in a `COPY INTO` statement.

5. What did you expect to see?

I expected to have the copy option set properly such that when this is compiled. `MAX_FILE_SIZE=50000000` appears in the generated SQL.

6. What did you see instead?

an error:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-44-783ee63d0133> in <module>
----> 1 unload_stmt.maxfilesize(50000000)

~/miniconda3/envs/civis36/lib/python3.6/site-packages/snowflake/sqlalchemy/custom_commands.py in maxfilesize(self, max_size)
    118     def maxfilesize(self, max_size):
    119         if not isinstance(max_size, bool):
--> 120             raise TypeError("Parameter max_size should  be a boolean value")
    121         self.copy_options.update({'MAX_FILE_SIZE': translate_bool(max_size)})
    122 

TypeError: Parameter max_size should  be a boolean value
```

7. Can you set logging to DEBUG and collect the logs?

```
import logging
import os

for logger_name in ['snowflake.sqlalchemy', 'snowflake.connector', 'botocore']: 
    logger = logging.getLogger(logger_name)
    logger.setLevel(logging.DEBUG)
    ch = logging.StreamHandler()
    ch.setLevel(logging.DEBUG)
[...]

PR #100, opened 2019-05-13
maxfilesize method on CopyInto checks for int type

colemandsmith:
This PR should address the [issue I encountered](https://github.com/snowflakedb/snowflake-sqlalchemy/issues/99) with MAX_FILE_SIZE. The changes can be summed up as follows:

- `maxfilesize` will check for `int` type and raise the type error with an appropriate message if the provided arg is not an int
- in `visit_copy_into`, for the `copy_options`, `_compiler_dispatch` is only called if the value has the method as an attribute; otherwise, str is called.
  - the pattern for this was taken from the method `visit_copy_formatter` below `visit_copy_into`, for consistency, though a helper method to perform this logic could be a useful pattern as well

- updates to the test `test_copy_into_location` to ensure that the MAX_FILE_SIZE can be set properly via `maxfilesize` and that the statement will compile correctly.

### DIFF
--- a/base.py
+++ b/base.py
@@ -191,8 +191,9 @@ class SnowflakeCompiler(compiler.SQLCompiler):
         options_list = list(copy_into.copy_options.items())
         if kw.get('deterministic', False):
             options_list.sort(key=operator.itemgetter(0))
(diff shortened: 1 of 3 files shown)


notkriswagner:
Merged this into the release branch

Issue #97, opened 2019-05-10
CopyInto statement doesn't compile correctly when the source is storage and the destination is a table


colemandsmith:
Please answer these questions before submitting your issue. Thanks!

1. What version of Python are you using (`python --version`)?

Python 3.6.8 :: Anaconda, Inc.

2. What operating system and processor architecture are you using (`python -c 'import platform; print(platform.platform())'`)?

Darwin-18.5.0-x86_64-i386-64bit

3. What are the component versions in the environment (`pip list`)? 

snowflake-connector-python   1.7.11  
snowflake-sqlalchemy               1.1.12

4. What did you do?
If possible, provide a recipe for reproducing the error.
A complete runnable program is good.

I was trying to use the CopyInto command in order to copy from a table to S3. From my understanding this should be possible since the `COPY INTO` syntax in Snowflake seems to go both ways (tables -> storage and storage -> tables). However, the table name becomes blank when trying to use this. As an example from your test suite:
```
food_items = Table("python_tests_foods", meta,
                       Column('id', Integer, Sequence('new_user_id_seq'), primary_key=True),
                       Column('name', String),
                       Column('quantity', Integer))
# fake key of course
storage = AWSBucket.from_uri('s3://backup').encryption_aws_sse_kms(
                                      '1234abcd-12ab-34cd-56ef-1234567890ab')
# this works
CopyInto(from_=food_items, into=storage, formatter=CSVFormatter()).compile(dialect=SnowflakeDialect()).string

# generates "COPY INTO 's3://foo/bar' FROM python_tests_foods FILE_FORMAT=(TYPE=csv) ENCRYPTION=(TYPE='AWS_SSE_KMS' KMS_KEY_ID='1234abcd-12ab-34cd-56ef-1234567890ab')"

# this does not work
CopyInto(from_=storage, into=food_items, formatter=CSVFormatter()).compile(dialect=SnowflakeDialect()).string

# generates 'COPY INTO  FROM (("\'s3://foo/bar\'", \'\', "ENCRYPTION=(TYPE=\'AWS_SSE_KMS\' KMS_KEY_ID=\'1234abcd-12ab-34cd-56ef-1234567890ab\')")) FILE_FORMAT=(TYPE=csv)'
```

5. What did you expect to see?

I would expect to see something like:
[...]

PR #101, opened 2019-05-13
CopyInto successfully compiles statement for storage to table

colemandsmith:
This PR should address the [issue I had with CopyInto](https://github.com/snowflakedb/snowflake-sqlalchemy/issues/97) for compiling statements where we `COPY INTO` a table *from* storage. The changes can be summarized as follows:
- Changed some of the setup logic in `visit_copy_into` in the `SnowflakeCompiler` so we can leave from_ alone if it is a table, call _compiler_dispatch if it is an AWSBucket or an AzureContainer, or format _compiler_dispatch into ‘({})’ if the argument is some other type (such as a SQLAlchemy select)
    - This logical change could potentially be abstracted out into a helper method to be reused wherever a similar check is done
- Added a case to the `test_copy_into_location` test for the case of having from_ as an AWSBucket and into_ as a table

### DIFF
--- a/base.py
+++ b/base.py
@@ -11,7 +11,7 @@ from sqlalchemy import util as sa_util
 from sqlalchemy.engine import default
 from sqlalchemy.schema import Table
 from sqlalchemy.sql import compiler, expression
-from sqlalchemy.sql.elements import quoted_name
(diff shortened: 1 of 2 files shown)


smtakeda:
#103 will include the change.

PR #131, opened 2019-11-16
Timestamp DDL renders wrong when precision value passed in

abhikalakuntla:
If field name is `TS` and precision is set to 6, the column definition will be rendered as `"TS" TIMESTAMP(6) WITHOUT TIME ZONE` 
when it should be `"TS" TIMESTAMP WITHOUT TIME ZONE(6)"`.


smtakeda:
Thanks for PR. Will merge this in the next release.
Do you have a test case? If so, can you provide it?

PR #136, opened 2019-12-12
Update get_columns to perform its own information schema query specific to a table

tchoedak:
The current implementation of the dialect's method `get_columns` reuses `_get_schema_columns` which runs a query for all columns against the current schema. This sometimes leads to errors, especially in schemas that have a large number of tables. I am currently observing intermittent errors  in my organization's schema containing over 700 tables.

The actual error is reported on snowflake's knowledge base:

https://community.snowflake.com/s/article/error-information-schema-query-returned-too-much-data-please-repeat-query-with-more-selective-predicates-when-querying-information-schema-view

A specific instance of this error appears as such:
```
snowflake.connector.errors.ProgrammingError: 090030 (22000): 0190d231-0117-469e-0000-18c980bbfd9a: Information schema query returned too much data.  Please repeat query with more selective predicates.
```

The solution implemented here is to enable this method to perform its own query specific to a single table instead of against all possible tables in the schema to retrieve column information while following all the same steps as the original `_get_schema_columns` so that column and type information is extracted correctly.

This issue was discovered when attempting to reflect the tables in this schema using techniques and approaches described here: https://docs.sqlalchemy.org/en/13/core/reflection.html#reflecting-database-objects

Direction/ Help Wanted:
* I'm not sure what the decorator @reflection.cache does. If this is important to decorate on functions that actually run information schema queries, I can add it to this fn as it now executes its own query.

### DIFF
--- a/snowdialect.py
+++ b/snowdialect.py
@@ -402,7 +402,6 @@ class SnowflakeDialect(default.DefaultDialect):
             connection, full_schema_name, **kw
         )
 
-        columns = []
         table_name = self.denormalize_name(table_name)
         current_table_pks = schema_primary_keys.get(table_name)
 
--- a/snowdialect.py
+++ b/snowdialect.py
@@ -391,11 +391,89 @@ class SnowflakeDialect(default.DefaultDialect):
         Gets all column info given the table info
         """
         schema = schema or self.default_schema_name
+        current_database, current_schema = self._current_database_schema(connection, **kw)
         if not schema:
-            _, schema = self._current_database_schema(connection, **kw)
+            schema = current_schema
 
-        schema_columns = self._get_schema_columns(connection, schema, **kw)
-        return schema_columns[self.normalize_name(table_name)]
+        full_schema_name = self._denormalize_quote_join(current_database, schema)
+        schema_primary_keys = self._get_schema_primary_keys(connection, full_schema_name, **kw)
+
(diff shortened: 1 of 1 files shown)


keller00:
The reason why we do it this way is performance, `reflection.cache` is SQLAlchemy's built-in cache. For now you could downgrade 1-2 releases to solve your issue, as the Python connector used to behave exactly the way you have coded it to work, but be ready for much worse inspector speeds.

I wasn't aware of this limit on results, and I'm going to have to investigate the problem more to find a solution that does not compromise inspector speeds.

keller00:
Please downgrade to `v1.1.16` while we figure out what to do.

tchoedak:
We've pinned to v.1.1.16 and can confirm we're no longer seeing the `query returned too much data` error. We'll stay tuned for a solution as we're likely to get this error again as soon as we upgrade/unpin.

keller00:
I have fixed the error. The next release will try schema wide column retrieval and if it fails with this specific error, then it will fall back to table wide. Both ways are cached and even the schema wide timeout will be cached.
So everything will be faster then the older versions and work properly.

Please, report back once the new version is released.

Thank you for the PR!

tchoedak:
@keller00 thanks that's great to hear. Looking forward to this update and seeing how you guys implemented it.

PR #137, opened 2019-12-12
fix handling of empty table comments

ms32035:
Current code returns an empty string, which causes alembic autogenerated migrations to break.

### DIFF
--- a/snowdialect.py
+++ b/snowdialect.py
@@ -501,7 +501,7 @@ class SnowflakeDialect(default.DefaultDialect):
                                     )
         cursor = connection.execute(sql_command)
         ans = cursor.fetchone()
-        return {'text': ans['comment']}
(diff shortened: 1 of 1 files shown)


anujkumar93:
Can confirm this works. Without this, I get an auto-generated migration, even when there's no changes to the schema:
```
op.drop_table_comment(
        'user',
        existing_comment='',
        schema=None
    )
```
I get this for each of my models in my flask app that I am using alembic to generate migrations for.

ms32035:
@smtakeda @keller00

keller00:
I'm sorry, trying to catch up 😅 
Do you think that Column comments should work the same way?
I was trying to see the SQLAlchemy docs for guidance, but I can't seem to find this case.

keller00:
I have verified the fix, we're accepting it and it will be included in the next release.
Thank you!

PR #140, opened 2019-12-13
Add `driver` property to SnowflakeDialect

### LINKED ISSUE #139: AttributeError: 'SnowflakeDialect' object has no attribute 'driver'

1. What version of Python are you using (`python --version`)?
3.7.3

2. What operating system and processor architecture are you using (`python -c 'import platform; print(platform.platform())'`)?
Linux 64bit

3. What are the component versions in the environment (`pip list`)? 
snowflake-sqlalchemy
petl

4. What did you do?
```
from sqlalchemy import create_engine

engine = create_engine(
    'snowflake://{user}:{password}@{account}/'.format(
        user='<your_user_login_name>',
        password='<your_password>',
        account='<your_account_name>',
    )
)

engine.driver
```

`snowflake`

6. What did you see instead?

`AttributeError: 'SnowflakeDialect' object has no attribute 'driver'`

---------------------------

Based on the [SQLAlchemy docs](https://docs.sqlalchemy.org/en/13/core/internals.html#sqlalchemy.engine.interfaces.Dialect), there should be a `driver` property on a `Dialect`, but the Snowflake dialect currently does not implement this.

aiguofer:
This PR fixes #139. 

SQLAlchemy [defined `Dialects`](https://docs.sqlalchemy.org/en/13/core/internals.html#sqlalchemy.engine.interfaces.Dialect) with a `driver` property. I wasn't sure what to name it. Since this is the official (and likely only implementation) for snowflake connections, I went with `snowflake`, but it could also be `default` or something. I'd be happy to change it if desired.

### DIFF
--- a/snowdialect.py
+++ b/snowdialect.py
@@ -71,6 +71,7 @@ ischema_names = {
 
 class SnowflakeDialect(default.DefaultDialect):
     name = 'snowflake'
+    driver = 'snowflake'
     max_identifier_length = 65535
     cte_follows_insert = True
 


snowstakeda:
Fixed in https://github.com/snowflakedb/snowflake-sqlalchemy/commit/dc92fd91

PR #141, opened 2019-12-15
Fix typo in README Connection Parameters

anujkumar93:
I tried to connect to snowflake using flask-sqlalchemy (that uses alembic) db setup, and the connection string that worked for me was the one corrected here. I think seeing how connection parameters are usually defined in URLs, this felt more like a typo in the official docs.


snowstakeda:
Fixed in https://github.com/snowflakedb/snowflake-sqlalchemy/commit/f3a20ff0

PR #149, opened 2020-01-15
Correctly handle spaces in connection string passwords

aparkerlue:
`snowflake.sqlalchemy.URL` currently calls `quote_plus` to `quote` passwords for the connection string, but this approach incorrectly handles spaces in passwords. Calling `quote` instead properly percent-encodes spaces.

### DIFF
--- a/test/test_unit_url.py
+++ b/test/test_unit_url.py
@@ -16,6 +16,9 @@ def test_url():
                password='test') == "snowflake://admin:test@testaccount/"
 
     assert URL(account='testaccount', user='admin',
+               password='test test') == "snowflake://admin:test%20test@testaccount/"
+
+    assert URL(account='testaccount', user='admin',
                password='test', database='testdb') == \
            "snowflake://admin:test@testaccount/testdb"
 
--- a/test/test_unit_url.py
+++ b/test/test_unit_url.py
@@ -16,6 +16,10 @@ def test_url():
                password='test') == "snowflake://admin:test@testaccount/"
 
     assert URL(account='testaccount', user='admin',
+               password='1-pass 2-pass 3-: 4-@ 5-/ 6-pass') == \
+           "snowflake://admin:1-pass 2-pass 3-%3A 4-%40 5-%2F 6-pass@testaccount/"
+
+    assert URL(account='testaccount', user='admin',
                password='test', database='testdb') == \
            "snowflake://admin:test@testaccount/testdb"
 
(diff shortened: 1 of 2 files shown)


keller00:
While this works, `quote_plus` also seems to work for me.
Did you have an issue, where it didn't work for you?

aparkerlue:
I found that `quote_plus` led to failed connections while `quote` succeeded for passwords that contain spaces. To be specific:

- Password spaces encoded as either raw spaces (i.e., ` `) or percent-encoded spaces (`%20`) do work in the connection string.
- Password spaces encoded as `+` in the connection string failed to connect.

Actually [SQLAlchemy contains code that addresses this question](https://github.com/sqlalchemy/sqlalchemy/blob/master/lib/sqlalchemy/engine/url.py):

```
>>> from sqlalchemy.engine.url import make_url
>>> str(make_url('snowflake://user:1-pass 2-pass 3-: 4-@ 5-/ 6-pass@hostname/'))
'snowflake://user:1-pass 2-pass 3-%3A 4-%40 5-%2F 6-pass@hostname/'
```

Moreover they provide functions from which to build a URL string:

```
>>> from sqlalchemy.engine.url import URL
>>> str(URL("snowflake", "user", "1-pass 2-pass 3-: 4-@ 5-/ 6-pass", "hostname"))
'snowflake://user:1-pass 2-pass 3-%3A 4-%40 5-%2F 6-pass@hostname'
```

Might a better approach be to rely on SQLAlchemy code to build the connection string?

aparkerlue:
`sqlalchemy.engine.url._rfc_1738_quote()` appears to be a private method, so I copied it into `util.py` and used it just for the password. I also updated the test case with the sample password in my previous comment.

It might make sense to do a more substantial overhaul of this function to call `sqlalchemy.engine.url.URL` for URL construction.

keller00:
That function definition hasn't changed for the last 6 years.
I would still just import it from SQLAlchemy, but I can make this change when I merge the PR.
Is that ok with you?

keller00:
Note that having a space at the end of the password still works for me with quote_plus, but it is probably not the right way to escape urls (even the doc says that `quote_plus` is for forms). So I do want to accept the PR.

aparkerlue:
That sounds good. My concern with importing a private function is that the function name could eventually change or the function could disappear altogether at some point, but as you point out, that seems unlikely.

Thank you. Please proceed.

aparkerlue:
That sounds good. My concern with importing a private function is that the function name could eventually change or the function could disappear altogether at some point, but as you point out, that seems unlikely.

Thank you. Please proceed.

aparkerlue:
@keller00—Did you decide not to merge the PR?

keller00:
We do not merge straight into master on this repo.
We merge into our internal repo manually and then sync it to the public one upon release.

PR #154, opened 2020-02-10
Allow get_table_comment to fetch view comments too

necaris:
Since SQLAlchemy recommends that views be reflected in the same ways as tables (see https://docs.sqlalchemy.org/en/13/core/reflection.html#reflecting-views), allow the `get_table_comment` method to return metadata for views too.

### DIFF
--- a/snowdialect.py
+++ b/snowdialect.py
@@ -583,7 +583,21 @@ class SnowflakeDialect(default.DefaultDialect):
                                     )
         cursor = connection.execute(sql_command)
         ans = cursor.fetchone()
-        return {'text': ans['comment'] if ans['comment'] else None}
+        if ans is None:
+            # NOTE: this may happen because the "table" being reflected is
+            # actually a view; SQLAlchemy has the same code path for both.
+            sql_command = "SHOW /* sqlalchemy:get_view_comment */ " \
+                "VIEWS LIKE '{0}'{1}".format(
+                    table_name,
+                    (' IN SCHEMA {0}'.format(self.normalize_name(schema))) if schema else ''
+                )
+            cursor = connection.execute(sql_command)
+            ans = cursor.fetchone()
+
+        # Also, we should not throw an error here just because we could not
+        # fetch the *comment* field
+        comment = ans.get('comment', None) if ans else None
+        return {'text': comment if comment else None}
 
 
 @sa_vnt.listens_for(Table, 'before_create')
--- a/snowdialect.py
+++ b/snowdialect.py
@@ -607,14 +607,14 @@ class SnowflakeDialect(default.DefaultDialect):
         """
         try:
             comment = self._get_table_comment()
-            if comment is None:
+            if comment is None or comment == "":
                 # the "table" being reflected is actually a view
                 comment = self._get_view_comment()
         except (KeyError, TypeError):
             # Also, we should not throw an error here just because we could not
             # fetch the *comment* field
             comment = None
-        return {'text': comment}
+        return {'text': comment if comment else None}
 
 
 @sa_vnt.listens_for(Table, 'before_create')
--- a/snowdialect.py
+++ b/snowdialect.py
@@ -607,7 +607,7 @@ class SnowflakeDialect(default.DefaultDialect):
         """
         try:
             comment = self._get_table_comment()
-            if comment is None or comment == "":
+            if comment is None:
                 # the "table" being reflected is actually a view
                 comment = self._get_view_comment()
         except (KeyError, TypeError):
--- a/snowdialect.py
+++ b/snowdialect.py
@@ -576,7 +576,7 @@ class SnowflakeDialect(default.DefaultDialect):
         """
         Returns comment of table in a dictionary as described by SQLAlchemy spec.
         """
-        sql_command = "SHOW /* sqlalchemy:get_table_comment */ " \
+        sql_command = "SHOW /* sqlalchemy:_get_table_comment */ " \
                       "TABLES LIKE '{0}'{1}".format(
                                         table_name,
                                         (' IN SCHEMA {0}'.format(self.normalize_name(schema))) if schema else ''
@@ -589,7 +589,7 @@ class SnowflakeDialect(default.DefaultDialect):
         """
         Returns comment of view in a dictionary as described by SQLAlchemy spec.
         """
-        sql_command = "SHOW /* sqlalchemy:get_view_comment */ " \
+        sql_command = "SHOW /* sqlalchemy:_get_view_comment */ " \
             "VIEWS LIKE '{0}'{1}".format(
                 table_name,
                 (' IN SCHEMA {0}'.format(self.normalize_name(schema))) if schema else ''
@@ -601,10 +601,10 @@ class SnowflakeDialect(default.DefaultDialect):
     def get_table_comment(self, connection, table_name, schema=None, **kw):
         """
         Returns comment associated with a table (or view) in a dictionary as
-        SQLAlchemy expects.
+        SQLAlchemy expects. Note that since SQLAlchemy may not (in fact,
+        typically does not) know if this is a table or a view, we have to
+        handle both cases here.
         """
-        # NOTE: Since SQLAlchemy may not (in fact, typically does not) know if
-        # this is a table or a view, we have to handle both cases here.
         try:
             comment = self._get_table_comment()
             if comment is None:
--- a/snowdialect.py
+++ b/snowdialect.py
@@ -596,8 +596,8 @@ class SnowflakeDialect(default.DefaultDialect):
 
         # Also, we should not throw an error here just because we could not
         # fetch the *comment* field
-        comment = ans.get('comment', None) if ans else None
-        return {'text': comment if comment else None}
+        comment = ans['comment'] if (ans and ans['comment']) else None
+        return {'text': comment}
 
 
 @sa_vnt.listens_for(Table, 'before_create')
--- a/snowdialect.py
+++ b/snowdialect.py
@@ -582,8 +582,7 @@ class SnowflakeDialect(default.DefaultDialect):
                                         (' IN SCHEMA {0}'.format(self.normalize_name(schema))) if schema else ''
                                     )
         cursor = connection.execute(sql_command)
-        result = cursor.fetchone()
-        return result['comment']
+        return cursor.fetchone()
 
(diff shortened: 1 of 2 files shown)


snowdialect.py
```diff
@@ -583,7 +583,21 @@ def get_table_comment(self, connection, table_name, schema=None, **kw):
...
+        if ans is None:
+            # NOTE: this may happen because the "table" being reflected is
+            # actually a view; SQLAlchemy has the same code path for both.
+            sql_command = "SHOW /* sqlalchemy:get_view_comment */ " \
+                "VIEWS LIKE '{0}'{1}".format(
+                    table_name,
+                    (' IN SCHEMA {0}'.format(self.normalize_name(schema))) if schema else ''
+                )
+            cursor = connection.execute(sql_command)
+            ans = cursor.fetchone()
```
keller00: Could you make this a separate function, like `get_view_comment`?

keller00: Please install `tox` and make sure `tox -e fix_lint` passes.

snowdialect.py
```diff
@@ -583,7 +583,21 @@ def get_table_comment(self, connection, table_name, schema=None, **kw):
...
+
+        # Also, we should not throw an error here just because we could not
+        # fetch the *comment* field
+        comment = ans.get('comment', None) if ans else None
+        return {'text': comment if comment else None}
```
keller00: This if statement is not necessary. I'm pretty sure you catch all corner cases on the previous line.

keller00: You resolved this before I finished commenting 😄

snowdialect.py
keller00: Actually, don't worry about the `fix_lint`, it's made to work on our internal repo. I'll deal with linting issues.

akdor1154:
ah I was just gonna PR for exactly this.. @necaris do you have time to update this to meet requirements or would you like me to pick this up?

necaris:
I'd have time to fix this up, what does it need?

keller00:
You could add a test, like how I have outlined and I also asked for your input on column comments.

necaris:
Thanks @keller00 I've been away from a keyboard this morning but I should be able to get to the points you mentioned tonight -- thanks for the review!

necaris:
@keller00 please note I've updated the PR and believe I've addressed all your comments, would appreciate another look. Note that I've factored `_get_view_comment` into a helper of its own, but since SQLAlchemy reflection doesn't have a distinct concept of reflecting views,  `get_table_comment` still needs to manage it.

Regarding column comments: since both `get_schema_columns` and `get_table_columns` use the [columns view](https://docs.snowflake.net/manuals/sql-reference/info-schema/columns.html) and the doc for that says that filtering on `table_name` also allows a view name to be used, I'd hope it's not a problem. However, I'm not sure how filtering by `table_schema` behaves in that view, so I don't know for sure.

snowdialect.py
```diff
@@ -572,18 +572,49 @@ def get_schema_names(self, connection, **kw):
...
         sql_command = "SHOW /* sqlalchemy:get_table_comment */ " \
                       "TABLES LIKE '{0}'{1}".format(
                                         table_name,
                                         (' IN SCHEMA {0}'.format(self.normalize_name(schema))) if schema else ''
                                     )
         cursor = connection.execute(sql_command)
-        ans = cursor.fetchone()
-        return {'text': ans['comment'] if ans['comment'] else None}
+        result = cursor.fetchone()
+        return result['comment']
```
keller00: This needs to be changed back.
There was another bug where the empty string comment would trick alembic into thinking that there's a comment.
We need to check `result['comment'] == ''` and then instead return `None`.

snowdialect.py
```diff
@@ -572,18 +572,49 @@ def get_schema_names(self, connection, **kw):
...
+        Returns comment of view in a dictionary as described by SQLAlchemy spec.
+        """
+        sql_command = "SHOW /* sqlalchemy:get_view_comment */ " \
+            "VIEWS LIKE '{0}'{1}".format(
+                table_name,
+                (' IN SCHEMA {0}'.format(self.normalize_name(schema))) if schema else ''
+            )
+        cursor = connection.execute(sql_command)
+        result = cursor.fetchone()
+        return result['comment']
```
keller00: Same needs to be done here

snowdialect.py
```diff
@@ -572,18 +572,49 @@ def get_schema_names(self, connection, **kw):
...
-        ans = cursor.fetchone()
-        return {'text': ans['comment'] if ans['comment'] else None}
+        result = cursor.fetchone()
+        return result['comment']
+
+    def _get_view_comment(self, connection, table_name, schema=None, **kw):
+        """
+        Returns comment of view in a dictionary as described by SQLAlchemy spec.
+        """
+        sql_command = "SHOW /* sqlalchemy:get_view_comment */ " \
```
keller00: The function name is missing an underscore here

snowdialect.py
```diff
@@ -572,18 +572,49 @@ def get_schema_names(self, connection, **kw):
 
         return [self.normalize_name(row[1]) for row in cursor]
 
-    def get_table_comment(self, connection, table_name, schema=None, **kw):
+    def _get_table_comment(self, connection, table_name, schema=None, **kw):
         """
-        Returns comment of table in a dictionary as described by SQLAlchemy spec
+        Returns comment of table in a dictionary as described by SQLAlchemy spec.
         """
         sql_command = "SHOW /* sqlalchemy:get_table_comment */ " \
```
keller00: Please add the missing underscore to the function name here

snowdialect.py
```diff
@@ -572,18 +572,49 @@ def get_schema_names(self, connection, **kw):
...
+    def get_table_comment(self, connection, table_name, schema=None, **kw):
+        """
+        Returns comment associated with a table (or view) in a dictionary as
+        SQLAlchemy expects.
+        """
+        # NOTE: Since SQLAlchemy may not (in fact, typically does not) know if
+        # this is a table or a view, we have to handle both cases here.
+        try:
+            comment = self._get_table_comment()
+            if comment is None:
```
keller00: Note that this logic will need to be updated once you change the return value checks.

snowdialect.py
```diff
@@ -572,18 +572,49 @@ def get_schema_names(self, connection, **kw):
...
+        result = cursor.fetchone()
+        return result['comment']
+
+    def get_table_comment(self, connection, table_name, schema=None, **kw):
+        """
+        Returns comment associated with a table (or view) in a dictionary as
+        SQLAlchemy expects.
+        """
+        # NOTE: Since SQLAlchemy may not (in fact, typically does not) know if
+        # this is a table or a view, we have to handle both cases here.
```
keller00: This might be nitpicky, but I think this could move into the doc string right above it 😄

test/test_core.py
```diff
@@ -552,6 +552,42 @@ def test_view_definition(engine_testaccount, db_parameters):
...
+    engine_testaccount.execute(text(sql).execution_options(
+        autocommit=True))
+    try:
+        inspector = inspect(engine_testaccount)
+        # NOTE: sqlalchemy doesn't have a way to get view comments specifically,
+        # but the code to get table comments should work for views too
+        assert inspector.get_table_comment(test_view_name) == {'text': comment_text}
+    finally:
+        engine_testaccount.execute(
+            "DROP TABLE IF EXISTS {0}".format(test_table_name))
```
keller00: I'm just curious why did you wrap the `sql` in text on line 571, but not here.

keller00:
Hey @necaris thanks again!
There are just some small things and the reintroduction of a bug that is not obvious. Once you have made these changes I'll submit it for testing and we can get it merged soon after.

test/test_core.py
necaris: @keller00 now that you mention it, that does look odd -- I had in fact just copied it from the function above. I'll update them both.

necaris:
@keller00 I think I've got all your comments, please have another look. Thank you!

snowdialect.py
necaris: @keller00 please note that I'm checking this value in `get_table_comment` now, rather than repeating the logic in both helper functions, so this should be taken care of.

snowdialect.py
necaris: @keller00 please note that I'm checking this value in `get_table_comment` now, rather than repeating the logic in both helper functions, so this should be taken care of.

snowdialect.py
```diff
@@ -572,18 +572,49 @@ def get_schema_names(self, connection, **kw):
...
+        """
+        try:
+            comment = self._get_table_comment()
+            if comment is None or comment == "":
+                # the "table" being reflected is actually a view
+                comment = self._get_view_comment()
+        except (KeyError, TypeError):
+            # Also, we should not throw an error here just because we could not
+            # fetch the *comment* field
+            comment = None
```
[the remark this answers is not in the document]
necaris: @keller00 I've found that calling e.g. `result['comment']` on an empty `result` row throws an error that SQLAlchemy _doesn't catch_, so it actually just breaks reflection -- hence wanting to guard against errors here and just return `None`.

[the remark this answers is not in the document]
keller00: You're right, but then this whole PR won't work.
When trying to retrieve a VIew comment it will throw an exception in `_get_table_comment` and then this will catch it and not try to retrieve the view comment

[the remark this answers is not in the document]
keller00: In light of this, maybe `return result` in the internal functions instead and check in this outer function `if result is None:`.
But I'd still remove this error catching.

snowdialect.py
```diff
@@ -572,18 +572,49 @@ def get_schema_names(self, connection, **kw):
...
+    def get_table_comment(self, connection, table_name, schema=None, **kw):
+        """
+        Returns comment associated with a table (or view) in a dictionary as
+        SQLAlchemy expects. Note that since SQLAlchemy may not (in fact,
+        typically does not) know if this is a table or a view, we have to
+        handle both cases here.
+        """
+        try:
+            comment = self._get_table_comment()
+            if comment is None or comment == "":
```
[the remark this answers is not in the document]
necaris: Updated! @keller00 thank you for the catch

keller00:
There were a few bugs, but I have fixed them and will be accepting this PR.
Watch out for this in the next release.
Thank you @necaris 🏅 !

akdor1154:
thanks @necaris !

necaris:
Thank you both for the help!

PR #162, opened 2020-03-04
Adds ExternalStage

sburns:
I'm trying to use the very helpful `CopyIntoStorage` custom command to copy results into an S3 bucket using an IAM role (can confirm the the assume role policy from snowflake to our account is correctly setup on this role).

Something as simple as:

```python
from snowflake.sqlalchemy import AWSBucket, CopyIntoStorage, CSVFormatter

bucket = AWSBucket.from_uri("s3://bucket").credentials(aws_role="arn:.....role/role_name")
copy_into = CopyIntoStorage(from_=statement, into=bucket, formatter=CSVFormatter())
with engine.connect() as connection:
    connection.execute(copy_into)
```

Cannot work because [the docs](https://docs.snowflake.net/manuals/sql-reference/sql/copy-into-location.html#additional-cloud-provider-parameters) state this nugget:

> Using an IAM role for credentials requires a named external stage. Accessing an S3 bucket URL directly in a COPY statement is not supported.

Therefore, I've added an `ExternalStage` class that compiles to `@[namespace.]stage_name[/path]`

This, locally (more on that later) works just fine for me:

```python
formatter = CSVFormatter().compression('gzip').field_delimiter('|')
stage = ExternalStage(name="stage_name", path="prefix/file")
copy_into = CopyIntoStorage(from_=s, into=stage, formatter=formatter)
with engine.connect() as connection:
    connection.execute(copy_into)
```

I have added a few statements to test in `tests/test_copy.py` but the parameters provided in `tests/README.rst` don't work and I just get a ton of connection failures. Is this supposed to happen?

### DIFF
--- a/custom_commands.py
+++ b/custom_commands.py
@@ -289,6 +289,19 @@ class PARQUETFormatter(CopyFormatter):
         return self
 
 
+class ExternalStage(ClauseElement):
+    """External Stage descriptor"""
+    __visit_name__ = "external_stage"
+
+    def __init__(self, name, path='', namespace=''):
+        self.name = name
+        self.path = '/{}'.format(path) if path and not path.startswith('/') else path
+        self.namespace = '{}.' if namespace and not namespace.endswith('.') else namespace
+
+    def __repr__(self):
+        return "@{}{}{}".format(self.namespace, self.name, self.path)
+
+
 class AWSBucket(ClauseElement):
     """AWS S3 bucket descriptor"""
     __visit_name__ = 'aws_bucket'
--- a/custom_commands.py
+++ b/custom_commands.py
@@ -293,10 +293,18 @@ class ExternalStage(ClauseElement):
     """External Stage descriptor"""
     __visit_name__ = "external_stage"
 
-    def __init__(self, name, path='', namespace=''):
+    @staticmethod
+    def prepare_namespace(namespace):
+        return "{}.".format(namespace) if not namespace.endswith(".") else namespace
+
+    @staticmethod
+    def prepare_path(path):
+        return "/{}".format(path) if not namespace.startswith("/") else path
+
+    def __init__(self, name, path=None, namespace=None):
         self.name = name
-        self.path = '/{}'.format(path) if path and not path.startswith('/') else path
-        self.namespace = '{}.' if namespace and not namespace.endswith('.') else namespace
+        self.path = self.prepare_path(path) if path else ""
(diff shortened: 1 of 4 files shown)


sburns:
> I have added a few statements to test in tests/test_copy.py but the parameters provided in tests/README.rst don't work and I just get a ton of connection failures. Is this supposed to happen?

Reading `scripts/install.sh`, it looks like these params are relatively secret and encoded in `./parameters.py.enc`. So technically we could use credentials to our own accounts but then we'd be burning our own credits. I can see right through your business plan :)

keller00:
Hey @sburns thank you for the PR!

> Reading `scripts/install.sh`, it looks like these params are relatively secret and encoded in `./parameters.py.enc`. So technically we could use credentials to our own accounts but then we'd be burning our own credits. I can see right through your business plan :)

Ohh no, you caught on....
Don't worry about it, I can run the tests for you for free! 😄 

I have written this `COPY INTO` code a while ago, I need to get back to speed on it 😅

custom_commands.py
```diff
@@ -289,6 +289,19 @@ def snappy_compression(self, comp):
...
 
 
+class ExternalStage(ClauseElement):
+    """External Stage descriptor"""
+    __visit_name__ = "external_stage"
+
+    def __init__(self, name, path='', namespace=''):
+        self.name = name
+        self.path = '/{}'.format(path) if path and not path.startswith('/') else path
+        self.namespace = '{}.' if namespace and not namespace.endswith('.') else namespace
```
keller00: I think these lines could be problematic, I can see users passing in `None` instead of `''` and then this String format will be incorrectly `/None` and `None.`

sburns: Yeah I was thinking about that too and settled on default empty strings for now but am totally up for suggestions.

For paths, if passed a string and it doesn't start with a slash, we need to prepend the slash. Vice versa for the namespace and "."

Either way, what's the neat way to document this? README.rst?

sburns:
Any chance at getting a new pair of eyes on this @keller00 @smtakeda ?

keller00:
My bad, @sburns !
I got caught up in something, it looked good to me when I looked. I’ll run our tests on it today!

keller00:
Don't worry, @sburns 
I have ran testing and found/fixed the NameError you just pushed. It is ready to be merged internally. 😉 
Thank you again!

keller00:
Merged, will be included in the next release

sburns:
Thanks for shepherding it through @keller00!

PR #176, opened 2020-05-18
add wss.sh

sfc-gh-stakeda:
Add WhiteSource scan


sfc-gh-stakeda:
will merge this to the mono repo later.

Issue #201, opened 2020-11-19
SNOW-227101: Datatime columns with timezone are created as TIMEZONE_NTZ


Nazarii:
Please answer these questions before submitting your issue. Thanks!

1. What version of Python are you using?
`Python 3.6.9`
2. What operating system and processor architecture are you using ?
`Linux-4.15.0-123-generic-x86_64-with-Ubuntu-18.04-bionic`
3. What are the component versions in the environment ? 
```Package                       Version                   Location
----------------------------- ------------------------- -----------------------
addict                        2.2.1
aiofiles                      0.4.0
aiopg                         0.16.0
apitoolbox                    0.10.0
appdirs                       1.4.4
asn1crypto                    1.2.0
astroid                       2.3.3
atomicwrites                  1.3.0
attrs                         19.1.0
azure-common                  1.1.23
azure-core                    1.9.0
azure-storage-blob            12.6.0
azure-storage-common          2.1.0
beautifulsoup4                4.7.1
boto3                         1.9.157
botocore                      1.12.157
cached-property               1.5.1
cachetools                    3.1.1
certifi                       2019.3.9
cffi                          1.12.3
chardet                       3.0.4
Click                         7.0
coverage                      4.5.3
croniter                      0.3.34
cryptography                  2.8
dataclasses                   0.6
dbfread                       2.0.7
decorator                     4.4.0
defusedxml                    0.6.0
dnspython                     1.16.0
docutils                      0.14
email-validator               1.0.4
entrypoints                   0.3
et-xmlfile                    1.0.1
fastapi                       0.52.0
flake8                        3.7.9
flake8-quotes                 3.2.0
freezegun                     0.3.11
future                        0.18.2
google-api-core               1.22.0
google-api-python-client      1.7.4
google-auth                   1.20.1
google-auth-httplib2          0.0.3
google-cloud-bigquery         1.26.1
google-cloud-bigquery-storage 1.0.0
google-cloud-core             1.4.1
google-crc32c                 0.1.0
google-resumable-media        0.7.1
googleapis-common-protos      1.52.0
grpcio                        1.31.0
gunicorn                      19.9.0
h11                           0.8.1
httplib2                      0.13.0
httptools                     0.0.13
idna                          2.8
ijson                         3.1.1
importlib-metadata            0.15
iniconfig                     1.1.1
isodate                       0.6.0
isort                         4.3.20
itsdangerous                  1.1.0
jdcal                         1.4.1
[...]

PR #202, opened 2020-11-19
SNOW-227101: fix to datetime columns being created without timezone

Nazarii:
[Original ticket.](https://github.com/snowflakedb/snowflake-sqlalchemy/issues/201)

Commit 28e8031 introduced a bug where all `sqlalchemy.DateTime(timezone=True)` columns were created as `DATETIME` in Snowflake which is an alias to `TIMESTAMP_NTZ`.
This commit reverts datetime logic back so if the timezone is defined in datetime column it will create `TIMESTAMP_TZ` and `TIMESTAMP_NTZ` otherwise.
This bug was introduced in [v1.1.18](https://github.com/snowflakedb/snowflake-sqlalchemy/commit/28e80317968b51da401803fb98515e5e70a35885)

### DIFF
--- a/src/snowflake/sqlalchemy/base.py
+++ b/src/snowflake/sqlalchemy/base.py
@@ -634,10 +634,10 @@ class SnowflakeTypeCompiler(compiler.GenericTypeCompiler):
         return "BINARY"
 
     def visit_datetime(self, type_, **kw):
-        return "datetime"
+        return self.visit_TIMESTAMP(type_, **kw)
 
     def visit_DATETIME(self, type_, **kw):
-        return "DATETIME"
+        return self.visit_TIMESTAMP(type_, **kw)
 
     def visit_TIMESTAMP_NTZ(self, type_, **kw):
         return "TIMESTAMP_NTZ"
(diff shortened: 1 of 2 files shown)


WeiGuHS:
@Nazarii do you know when this commit can be merged? and also is a reason that take out these TIMESTAMP_TZ from v1.1.8?

Nazarii:
@WeiGuHS no idea, this PR has no reviewers assigned, the same with the original ticket. I have no permissions to assign PR, I'll try to ping other contributors to review. It looks like the support of `TIMESTAMP_TZ` was removed by a mistake.

Nazarii:

Nazarii:
Not sure why but this PR shows there are conflicts. Diff does not show any merge conflicts or maybe there are some other conflicts?

sfc-gh-mkeller:
We will revisit this issue shortly.
In the meantime `TIMESTAMP` is an alias that is configurable actually, please see [our docs](https://docs.snowflake.com/en/sql-reference/parameters.html#timestamp-type-mapping) to see how to change what it points at.

philippeboyd:
@sfc-gh-mkeller any news on this?

conflicts comes from GEOGRAPHY being added at the same spot in the file as your modifications

```python
    def visit_GEOGRAPHY(self, type_, **kw):
        return "GEOGRAPHY"
```

sfc-gh-mraba:
Closing — superseded by #649 ("Add support for timestamps with timezone"). On `main`, timezone-aware `DateTime`/`DATETIME` render as Snowflake's `TIMESTAMP_TZ` (and `TIMESTAMP_NTZ`/`LTZ` are supported), rather than the `TIMESTAMP WITH TIME ZONE` form proposed here. Thanks @Nazarii.

PR #209, opened 2021-01-14
SNOW-271500 Update dialect to wrap raw SQL for SQLA 1.4

TRManderson:
SQLAlchemy 1.4.0b1 throw `sqlalchemy.exc.ObjectNotExecutableError` any time `connection.execute` is called using an `AsyncConnection` on raw string SQL. This wraps the raw string SQL objects in `sqlalchemy.text` and updates the bind parameter syntax correspondingly to avoid this issue.

### DIFF
--- a/snowdialect.py
+++ b/snowdialect.py
@@ -228,6 +228,8 @@ class SnowflakeDialect(default.DefaultDialect):
 
     @reflection.cache
     def _current_database_schema(self, connection, **kw):
+        # "branches" a connection if one exists, otherwise creates one
+        # branched connections use the same underlying DBAPI but maintain separate references
         with connection.connect() as sqla_con:
             dbapi_con = sqla_con.connection
             return (
--- a/snowdialect.py
+++ b/snowdialect.py
@@ -228,10 +228,11 @@ class SnowflakeDialect(default.DefaultDialect):
 
     @reflection.cache
     def _current_database_schema(self, connection, **kw):
-        con = connection.connection
-        return (
-            self.normalize_name(con.database),
-            self.normalize_name(con.schema))
+        with connection.connect() as sqla_con:
+            dbapi_con = sqla_con.connection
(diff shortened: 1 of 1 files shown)


snowdialect.py
```diff
@@ -227,7 +228,7 @@ def _denormalize_quote_join(self, *idents):
 
     @reflection.cache
     def _current_database_schema(self, connection, **kw):
-        con = connection.connect().connection
+        con = connection.connection
```
sfc-gh-mkeller, line 231: I'm sorry, but this line's change doesn't pass our automated tests.
We have a testcase where this gets called where no connection is open yet.

I checked https://docs.sqlalchemy.org/en/13/core/connections.html#sqlalchemy.engine.Engine.connect and using it here seems correct since it will reuse a previously created connection if it's available. Do you think that it was wrongly used here?

sfc-gh-mkeller (changes requested):
Other than the one issue (I rolled the one line back locally) and then the tests are passing. So once we figure out the one line I'm okay to merge this!

snowdialect.py
TRManderson: I think this change comes from me caring about [Connection.connect](https://docs.sqlalchemy.org/en/13/core/connections.html#sqlalchemy.engine.Connection.connect) instead of Engine.connect, but even the code change there is fine.

I'm going to rewrite this section so that it makes use of the existing `connection.connect().connection` process but uses the `Connection` context manager ensure the branched `Connection` is properly closed (which doesn't actually do anything in real terms, just ensures cleanup happens ASAP rather than relying on the GC to remove the dangling connection)

TRManderson: Change pushed, hope that's suitable :)

sfc-gh-mkeller: Sounds good! I'll kick off tests 🚀

Issue #194, opened 2020-10-01
SNOW-201006: Table Meta-Data with Geography Type throwing error "SAWarning: Did not recognize type 'GEOGRAPHY' of column"


jpg013:
Please answer these questions before submitting your issue. Thanks!

1. What version of Python are you using (`python --version`)? 
3.6.11

2. What operating system and processor architecture are you using (`python -c 'import platform; print(platform.platform())'`)?
Linux-5.4.0-7642-generic-x86_64-with-debian-bullseye-sid

3. What are the component versions in the environment (`pip list`)?
asn1crypto                 1.4.0
azure-common               1.1.25
azure-core                 1.8.1
azure-storage-blob         12.5.0
boto3                      1.14.63
botocore                   1.17.63
certifi                    2020.6.20
cffi                       1.14.3
chardet                    3.0.4
cryptography               2.9.2
docopt                     0.6.2
docutils                   0.15.2
flake8                     3.8.3
GeoAlchemy2                0.8.4
idna                       2.9
importlib-metadata         2.0.0
isodate                    0.6.0
jmespath                   0.10.0
mccabe                     0.6.1
msrest                     0.6.19
oauthlib                   3.1.0
oscrypto                   1.2.1
pip                        20.2.3
psycopg2                   2.8.6
pycodestyle                2.6.0
pycparser                  2.20
pycryptodomex              3.9.8
pyflakes                   2.2.0
PyJWT                      1.7.1
pyOpenSSL                  19.1.0
python-dateutil            2.8.1
pytz                       2020.1
requests                   2.23.0
requests-oauthlib          1.3.0
s3transfer                 0.3.3
setuptools                 40.6.2
setuptools-scm             4.1.2
Shapely                    1.7.1
six                        1.15.0
snowflake-connector-python 2.3.2
snowflake-sqlalchemy       1.2.3
SQLAlchemy                 1.3.19
urllib3                    1.25.10
zipp                       3.2.0

4. What did you do?
If possible, provide a recipe for reproducing the error.
A complete runnable program is good.
I'm trying to retrieve the table meta-data for a Snowflake schema.table. See following code:
    md = MetaData(bind=connection) # connection is snowflake connection
    table = Table(
        table_name,
        md,
        schema=schema_name,
        autoload=True,
        autoload_with=connection
    )
    columns = table.c
    return columns

[...]

PR #215, opened 2021-02-15
Fix fallback to NullType when dialect type not found

Rustem:
GEOGRAPHY custom type is not supported yet. Decided to fix issue similarly how sqlalchemy/postgresql dialect fixes it using sqlalchemy generic type called NULLTYPE. For more information [see here](https://github.com/zzzeek/sqlalchemy/blob/master/lib/sqlalchemy/dialects/postgresql/base.py#L3753)

Root cause: `sqltypes.NULLTYPE` is already an instance of sqltypes.NullType, however when type is not supported logic does not account for that and tries to initalize already initialized instance.

**Related Issue**: https://github.com/snowflakedb/snowflake-sqlalchemy/issues/194

### DIFF
--- a/snowdialect.py
+++ b/snowdialect.py
@@ -380,8 +380,7 @@ class SnowflakeDialect(default.DefaultDialect):
                                 (sqltypes.String, sqltypes.BINARY)):
                     col_type_kw['length'] = character_maximum_length
 
-            type_instance = col_type(**col_type_kw)
-
+            type_instance = col_type if isinstance(col_type, sqltypes.NullType) else col_type(**col_type_kw)
             current_table_pks = schema_primary_keys.get(table_name)
 
             ans[table_name].append({


Rustem:
Ok This workaround does not work , just tested on a real instance

Rustem:
Could you help to merge this bug fix asap please? As we temporarily patched the dialect on our end.

Mafioso:
We have same problem with dialect. Can you please merge this pull request. @sfc-gh-kwagner @sfc-gh-mkeller

snowdialect.py
```diff
@@ -380,8 +380,7 @@ def _get_schema_columns(self, connection, schema, **kw):
                                 (sqltypes.String, sqltypes.BINARY)):
                     col_type_kw['length'] = character_maximum_length
 
-            type_instance = col_type(**col_type_kw)
-
+            type_instance = col_type if isinstance(col_type, sqltypes.NullType) else col_type(**col_type_kw)
```
Fraznist, line 383: We probably need the same change at line 453

Fraznist:
Any plans to merge this? This looks like a good hotfix to integrate. `engine.dialect.get_columns()` is unusable due to this if any table in schema contains a GEOGRAPHY column.@sfc-gh-kwagner @sfc-gh-mkeller

nikhiltjoy-dd:
+1 GEOGRAPHY columns are causing issues

hadia206:
+1 for merge request.
I am getting this error as well when trying to do `df.to_sql` if any table in schema contains a GEOGRAPHY column.

njriasan:
> +1 for merge request.
> I am getting this error as well when trying to do `df.to_sql` if any table in schema contains a GEOGRAPHY column.

+1 on the merge request. I have verified that applying this change resolves my issues.

sfc-gh-mraba:
Closing — implemented on `main`. The reflection type mapping now returns `NullType()` directly for unknown/unmappable column types (with a warning) instead of calling `col_type(**kwargs)`, so the crash this PR fixed no longer occurs. Thanks @Rustem — the guard you added is in place.

PR #245, opened 2021-10-06
Add JSON access query support

rickpr:
When attempting to query JSON using this syntax, I receive the following error:
``` python
sqlalchemy.exc.UnsupportedCompilationError: Compiler <snowflake.sqlalchemy.base.SnowflakeCompiler object at 0x13f6cefd0> can't render element of type <function json_getitem_op at 0x1103c4790> (Background on this error at: http://sqlalche.me/e/13/l7de)
```

This means my workaround is to do something like this:
``` python
session.query(table1.metadata.op(':')(sqlalchemy.text('size')))
```

This is not only ugly but also means that I can't prototype my operations using another database like PostgreSQL before using my code live on snowflake.

This makes JSON access possible using the regular SQLAlchemy syntax.


masamitsu-murase:
Will this pull request be merged into the main branch?
I would like to know the future plan regarding the support of JSON.

I also developed a small library, [snowflake-sqlalchemy-json](https://github.com/masamitsu-murase/snowflake-sqlalchemy-json) to support JSON and `func.flatten(...)`, but I'm not sure whether creating a pull request based on this library will help this project.
If someone knows the future plan about JSON, could you share it with me?

Regards, 
Murase

PR #272, opened 2022-02-25
SNOW-576659 fix dialect identifier max length and relevant schema introspection

saulbein:
The snowflake dialect has identifier max length of 255 according to [this page of the official documentation](https://docs.snowflake.com/en/sql-reference/identifiers.html). However, in this library, the limit was set to 65535. This causes issues when trying to use an sqlalchemy [naming convention](https://docs.sqlalchemy.org/en/14/core/constraints.html?highlight=naming%20convention#configuring-constraint-naming-conventions), because while sqlalchemy would take into account the identifier length and shorten long names where needed, the higher limit will make it think that there is no shortening needed. Without the shortening logic, sqlalchemy will emit queries that will error out because of hitting the identifier limts.

This PR updates the limit, adds a test so that the naming convention usecase is covered in the future and fixes some constraint introspection code to make it easier to write such tests.

### DIFF
--- a/snowdialect.py
+++ b/snowdialect.py
@@ -4,6 +4,7 @@
 # Copyright (c) 2012-2019 Snowflake Computing Inc. All right reserved.
 #
 
+from collections import defaultdict
 import operator
 from functools import reduce
 
@@ -89,7 +90,7 @@ ischema_names = {
 class SnowflakeDialect(default.DefaultDialect):
     name = 'snowflake'
     driver = 'snowflake'
-    max_identifier_length = 65535
+    max_identifier_length = 255
     cte_follows_insert = True
 
     # TODO: support SQL caching, for more info see: https://docs.sqlalchemy.org/en/14/core/connections.html#caching-for-third-party-dialects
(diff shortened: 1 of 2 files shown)


saulbein:

saulbein:
recheck

snowdialect.py
```diff
@@ -296,6 +301,41 @@ def get_pk_constraint(self, connection, table_name, schema=None, **kw):
...
+                    'table_name': self.normalize_name(row['table_name'])
+                }
+            else:
+                unique_constraints[name]['column_names'].append(self.normalize_name(row['column_name']))
+
+        ans = {}
+        for _, v in iteritems(unique_constraints):
+            if v['table_name'] not in ans:
+                ans[v['table_name']] = []
+            ans[v['table_name']].append({k2: v2 for k2, v2 in iteritems(v) if k2 != 'table_name'})
```
sfc-gh-mkeller, line 325: Just as an FYI I think this would be a little more concise
```suggestion
        for _, v in iteritems(unique_constraints):
            ans.setdefault(v['table_name'], list()).append({k2: v2 for k2, v2 in iteritems(v) if k2 != 'table_name'})
```

sfc-gh-mkeller: This is my minimal version, but I could have added some bugs. Please see if you liked this, I think this is easier to understand (but I know that's relative).
```suggestion
        for _, v in iteritems(unique_constraints):
            table_name = v.pop('table_name')
            ans.setdefault(v[table_name], list()).append(v)
```

sfc-gh-mkeller (changes requested):

snowdialect.py
saulbein: To be honest, I copied these from `_get_schema_foreign_keys`, but if we're cleaning things up, I updated to what I would consider the cleanest version :slightly_smiling_face:

sfc-gh-mkeller (approved):

PR #273, opened 2022-03-10
Fix comment dropping operations

saulbein:
SQLAlchemy operations for dropping table / column comments were failing because the default dialect implementation generates these commands:
```sql
COMMENT ON TABLE table1 IS NULL
COMMENT ON COLUMN table1.id IS NULL
```
Snowflake only accepts strings for comment values, so the queries failed with syntax errors.

I've also added schema to a compiler test table to verify that it creates correct statements with a non-default schema.

### DIFF
--- a/test/test_compiler.py
+++ b/test/test_compiler.py
@@ -18,7 +18,7 @@ table2 = table(
     column('id', Integer),
     column('name', String),
     column('value', Integer),
-    schema="test"
+    schema="test",
 )
 
 
--- a/test/test_compiler.py
+++ b/test/test_compiler.py
@@ -2,6 +2,7 @@
 # -*- coding: utf-8 -*-
 
 from sqlalchemy import Integer, String, and_, select
+from sqlalchemy.schema import DropColumnComment, DropTableComment
 from sqlalchemy.sql import column, quoted_name, table
 from sqlalchemy.testing import AssertsCompiledSQL
 
@@ -16,7 +17,8 @@ table2 = table(
     'table2',
     column('id', Integer),
     column('name', String),
-    column('value', Integer)
+    column('value', Integer),
+    schema="test"
 )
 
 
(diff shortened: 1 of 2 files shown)


test/test_compiler.py
```diff
@@ -16,7 +17,8 @@
     'table2',
     column('id', Integer),
     column('name', String),
-    column('value', Integer)
+    column('value', Integer),
+    schema="test"
```
sfc-gh-mkeller, line 21:
```suggestion
    schema="test",
```

sfc-gh-mkeller:
This looks good to me @saulbein but there's currently no way in `snowflake-sqlalchemy` to add a comment to tables, or columns. I think we should only add dropping comments if they can be added through the library as well.

Would you be able to submit code for that too?

saulbein:
The SQLAlchemy default dialect has support for both adding and dropping comments, I was only overriding the drop operations as those by default set the comment to `NULL`, which emits queries that do not work (I found this out by running alembic migrations). I can override the others as well if that makes the code nicer, but they are already working.

sfc-gh-mkeller:
The reason I ask was because I was hoping to reuse the code for setting a comment for dropping comments. In essence 

```
def visit_drop_table_comment():
    return self.visit_table_comment("")
```

Do you think that we could do something like this?

saulbein:
I checked and it is possible to re-map to `visit_set_table_column`, but only if we allow `Table` objects, instead of `TableClause` elements that core `table()` function creates. I feel like limiting comment operations to only ORM tables is worse than the additional maintenance cost of having another drop statement.

As for the column comments, if we switch to `COMMENT ON COLUMN table.column IS ''` then reflection will not work well, as it currently returns `None` for columns without comments and would return `''` instead. For table columns this is dealt with in reflection already.

saulbein:
Actually, the current `SetColumnComment` operation already only works with ORM `Table`s, so nothing would be lost. Will update the code.

sfc-gh-mkeller:
Okay, that makes sense and thanks for the PR!

sfc-gh-mkeller (approved):

PR #277, opened 2022-04-06
Set foreign key referred schema to None when needed

saulbein:
SQLAlchemy expects foreign key reflection to set schemas which do not need specifying in the current database context (i.e. default schema or current schema) to be set to None. Without this, alembic migration keep re-creating foreign keys as it considers a referred table with a schema and without a schema two separate tables.

See [SQLAlchemy docs on reflection](https://docs.sqlalchemy.org/en/14/core/reflection.html#reflection-schema-qualified-interaction) for what is the expected behaviour of dialects.

### DIFF
--- a/snowdialect.py
+++ b/snowdialect.py
@@ -298,6 +298,7 @@ class SnowflakeDialect(default.DefaultDialect):
 
     @reflection.cache
     def _get_schema_foreign_keys(self, connection, schema, **kw):
+        _, current_schema = self._current_database_schema(connection, **kw)
         result = connection.execute(text(
             "SHOW /* sqlalchemy:_get_schema_foreign_keys */ IMPORTED KEYS IN SCHEMA {0}".format(schema)
         ))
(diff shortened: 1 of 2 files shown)


snowdialect.py
```diff
@@ -298,16 +298,22 @@ def get_pk_constraint(self, connection, table_name, schema=None, **kw):
...
             name = self.normalize_name(row['fk_name'])
             if name not in foreign_key_map:
+                referred_schema = self.normalize_name(row['pk_schema_name'])
                 foreign_key_map[name] = {
                     'constrained_columns': [self.normalize_name(row['fk_column_name'])],
-                    'referred_schema': self.normalize_name(row['pk_schema_name']),
+                    # referred schema should be None in context where it doesn't need to be specified
+                    # https://docs.sqlalchemy.org/en/14/core/reflection.html#reflection-schema-qualified-interaction
+                    'referred_schema': (referred_schema
+                                        if referred_schema not in [self.default_schema_name, current_schema]
```
sfc-gh-mkeller, line 315: Tuples in general are a little faster and they're immutable
```suggestion
                                        if referred_schema not in (self.default_schema_name, current_schema)
```

sfc-gh-mkeller (changes requested):

sfc-gh-mkeller (approved):

sfc-gh-mkeller:
I hope you don't mind that I just pushed my small change into your PR, but other than that tiny change this looks good

Issue #124, opened 2019-10-24
insert with autoincrement fails with sqlalchemy.exc.ArgumentError


ynux:
## What did I do

Create a table with an autoincrement  column as in your [README](https://github.com/snowflakedb/snowflake-sqlalchemy#auto-increment-behavior) . Then fetch the table into a new MetaData object and insert, relying on autoincrement / no value for that column.

### What did I expect

A row to be created, with the autoincrement column filled from the sequence

### Error

sqlalchemy.exc.ArgumentError: Column type DECIMAL(38, 0) on column 'sequencetable.id' is not compatible with autoincrement=True

### Reproduce

```
metadata = MetaData()
t1 = Table('sequencetable', metadata,
    Column('id', Integer, Sequence('id_seq'), primary_key=True),
    Column('data', String(39))
)

t1.create(engine)

### this works
connection.execute(t1.insert(), [ {'data': 'test_insert'}])

select_stmt_t1 = select([t1])
connection.execute(select_stmt_t1).fetchall()

metadata2 = MetaData()
t2 = Table('sequencetable', metadata2, autoload=True, autoload_with=engine)

### this fails 
connection.execute(t2.insert(), [ {'data': 'multi_insert_1'}, {'data': 'multi_insert_2'} ])
connection.execute(t2.insert(), [ {'data': 'test_insert'}])

### this works
seq = Sequence('id_seq')
nextid = connection.execute(seq)
connection.execute(t2.insert(), [ {'id': nextid, 'data': 'test_insert'}])

```
Remark: The above code stopped working for me now. When I create the table, it fails with 

```
sqlalchemy.exc.ProgrammingError: (snowflake.connector.errors.ProgrammingError) 090106 (22000): Cannot perform CREATE SEQUENCE. This session does not have a current schema. Call 'USE SCHEMA', or use a qualified name.
[SQL: CREATE SEQUENCE id_seq]
```
As a workaround I will create and use the sequence explicitly.

While you are at it, maybe correct the `autoinrecment` typo in your tests.

### Version info
Python 3.7.4
Darwin-18.7.0-x86_64-i386-64bit
```
Package                    Version  
-------------------------- ---------
asn1crypto                 1.2.0    
azure-common               1.1.23   
azure-storage-blob         2.1.0    
azure-storage-common       2.1.0    
boto3                      1.9.253  
botocore                   1.12.253 
certifi                    2019.9.11
cffi                       1.13.1   
[...]

PR #297, opened 2022-06-01
SNOW-118945: Fix ArgumentError where insert with autoincrement failed due to incompatible column type affinity

sfc-gh-aling:

   Fixes SNOW-118945, #124 

The code is not doing correct in **_two_** aspects:

1. The first issue is that when we create a column with `Sequence`, the default value was not set correctly, it should not be set using "autoincrement" which failed to connect the column with the sequence.
       - to fix the issue, we should use "default seq.nextval" to set as the column's default value.
       - for more context please check doc here for "autoincrement" and "default" keyword: https://docs.snowflake.com/en/sql-reference/sql/create-table.html#optional-parameters
2. The second issue is caused by [type affinity check](https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/sql/schema.py#L4504-L4511), specifically sqlalchemy is checking whether DECIMAL(NUMBER in snowflake) is a subclass of INTEGER (this is what user hit).
       - In snowflake, INTEGER is synonymous of NUMERIC (DECIMAL), see doc here: https://docs.snowflake.com/en/sql-reference/data-types-numeric.html, when we create a sqlalchemy `Integer` in snowflake, a NUMBER(38, 0) (DECIMAL is sqlalchemy) is created.
       - to fix the issue, we need to customize the affinity check of DECIMAL in which case DECIMAL(38,0) is affinity of INTEGER.

### DIFF
--- a/custom_types.py
+++ b/custom_types.py
@@ -57,4 +57,4 @@ class GEOGRAPHY(SnowflakeType):
 class _CUSTOM_DECIMAL(SnowflakeType, sqltypes.DECIMAL):
     @util.memoized_property
     def _type_affinity(self):
-        return sqltypes.INTEGER if self.precision == 38 and self.scale == 0 else sqltypes.DECIMAL
+        return sqltypes.INTEGER if self.scale == 0 else sqltypes.DECIMAL
--- a/custom_types.py
+++ b/custom_types.py
@@ -5,6 +5,7 @@
 #
 
 import sqlalchemy.types as sqltypes
+import sqlalchemy.util as util
 
 TEXT = sqltypes.VARCHAR
 CHARACTER = sqltypes.CHAR
@@ -51,3 +52,9 @@ class TIMESTAMP_NTZ(SnowflakeType):
 
 class GEOGRAPHY(SnowflakeType):
     __visit_name__ = 'GEOGRAPHY'
+
+
+class _CUSTOM_DECIMAL(SnowflakeType, sqltypes.DECIMAL):
+    @util.memoized_property
+    def _type_affinity(self):
+        return sqltypes.INTEGER if self.precision == 38 and self.scale == 0 else sqltypes.DECIMAL
--- a/snowdialect.py
+++ b/snowdialect.py
@@ -45,7 +45,7 @@ from .base import (
     SnowflakeIdentifierPreparer,
     SnowflakeTypeCompiler,
 )
-from .custom_types import ARRAY, GEOGRAPHY, OBJECT, TIMESTAMP_LTZ, TIMESTAMP_NTZ, TIMESTAMP_TZ, VARIANT
+from .custom_types import _CUSTOM_DECIMAL, ARRAY, GEOGRAPHY, OBJECT, TIMESTAMP_LTZ, TIMESTAMP_NTZ, TIMESTAMP_TZ, VARIANT
 
 colspecs = {}
 
@@ -65,7 +65,7 @@ ischema_names = {
     'FLOAT': FLOAT,
     'INT': INTEGER,
     'INTEGER': INTEGER,
-    'NUMBER': DECIMAL,
+    'NUMBER': _CUSTOM_DECIMAL,
     # 'OBJECT': ?
     'REAL': REAL,
     'BYTEINT': SMALLINT,
@@ -609,7 +609,7 @@ class SnowflakeDialect(default.DefaultDialect):
             ret = cursor.fetchone()
             if ret:
                 return ret[n2i['text']]
-        except Exxception:
+        except Exception:
             pass
         return None
 
--- a/test/test_quote.py
+++ b/test/test_quote.py
@@ -11,7 +11,7 @@ def test_table_name_with_reserved_words(engine_testaccount, db_parameters):
     metadata = MetaData()
     test_table_name = 'insert'
     insert_table = Table(test_table_name, metadata,
-                         Column('id', Integer, Sequence(test_table_name + '_id_seq'),
+                         Column('id', Integer, Sequence(f"{test_table_name}_id_seq"),
                                 primary_key=True),
                          Column('name', String),
                          Column('fullname', String),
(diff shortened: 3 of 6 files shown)


test/test_quote.py
```diff
@@ -22,10 +22,10 @@ def test_table_name_with_reserved_words(engine_testaccount, db_parameters):
         inspector = inspect(engine_testaccount)
         columns_in_insert = inspector.get_columns(test_table_name)
         assert len(columns_in_insert) == 3
-        assert columns_in_insert[0]['autoincrement'], 'autoincrement'
-        assert columns_in_insert[0]['default'] is None, 'default'
-        assert columns_in_insert[0]['name'] == 'id', 'name'
-        assert columns_in_insert[0]['primary_key'], 'primary key'
+        assert columns_in_insert[0]['autoincrement'] is False
```
sfc-gh-aling, line 25: for more context on the test change, `autoincrement` and `default` are mutually exclusive in snowflake according to [doc](https://docs.snowflake.com/en/sql-reference/sql/create-table.html#optional-parameters).

by fixing the `sequence` default value issue, we need to update this test accordingly.

custom_types.py
```diff
@@ -51,3 +52,9 @@ class TIMESTAMP_NTZ(SnowflakeType):
 
 class GEOGRAPHY(SnowflakeType):
     __visit_name__ = 'GEOGRAPHY'
+
+
+class _CUSTOM_DECIMAL(SnowflakeType, sqltypes.DECIMAL):
+    @util.memoized_property
+    def _type_affinity(self):
+        return sqltypes.INTEGER if self.precision == 38 and self.scale == 0 else sqltypes.DECIMAL
```
sfc-gh-mkeller, line 60: I think precision is not important here, `NUMBER(X, 0)` should always be an integer

snowdialect.py
```diff
@@ -609,7 +609,7 @@ def get_view_definition(self, connection, view_name, schema=None, **kw):
             ret = cursor.fetchone()
             if ret:
                 return ret[n2i['text']]
-        except Exxception:
```
sfc-gh-mkeller, line 612: Oh, how did we not catch this until now? O.o

test/test_sequence.py
```diff
@@ -0,0 +1,84 @@
...
+    seq = Sequence(test_table_name + '_id_seq')
+    try:
+        engine_testaccount.execute(sequence_table.insert(), [{'data': 'test_insert_1'}])
+
+        select_stmt = select([sequence_table])
+        result = engine_testaccount.execute(select_stmt).fetchall()
+        assert result == [(1, 'test_insert_1')]
+
+        metadata_autoload = MetaData()
+        autoload_sequence_table = Table(test_table_name, metadata_autoload, autoload=True, autoload_with=engine_testaccount)
```
sfc-gh-mkeller, line 28: `metadata_autoload` is never used again, you could inline it here

test/test_sequence.py
```diff
@@ -0,0 +1,84 @@
...
+
+        engine_testaccount.execute(autoload_sequence_table.insert(),
+                                   [{'data': 'multi_insert_1'}, {'data': 'multi_insert_2'}])
+
+        engine_testaccount.execute(autoload_sequence_table.insert(), [{'data': 'test_insert_2'}])
+
+        nextid = engine_testaccount.execute(seq)
+        engine_testaccount.execute(autoload_sequence_table.insert(), [{'id': nextid, 'data': 'test_insert_seq'}])
+        result = engine_testaccount.execute(select_stmt).fetchall()
```
sfc-gh-mkeller, line 37: Since you check list to another list a `ORDER BY` clause could prevent some falkiness in the future.

test/test_sequence.py
```diff
@@ -0,0 +1,84 @@
...
+        assert result == [
+            (1, 'test_insert_1'),
+            (2, 'multi_insert_1'),
+            (3, 'multi_insert_2'),
+            (4, 'test_insert_2'),
+        ]
+    finally:
+        autoincrement_table.drop(engine_testaccount)
+    return autoincrement_table
```
sfc-gh-mkeller, line 84:
```suggestion
```

test/test_sequence.py
```diff
@@ -0,0 +1,84 @@
...
+            (1, 'test_insert_1'),
+            (2, 'multi_insert_1'),
+            (3, 'multi_insert_2'),
+            (4, 'test_insert_2'),
+            (5, 'test_insert_seq')
+        ]
+    finally:
+        sequence_table.drop(engine_testaccount)
+        seq.drop(engine_testaccount)
+    return sequence_table
```
sfc-gh-mkeller, line 48: Why does your test return anything?
```suggestion
```

sfc-gh-mkeller (changes requested):

custom_types.py
sfc-gh-aling: aha, agreed, code updated

snowdialect.py
sfc-gh-aling: maybe no one has ever met with the exception case, I was surprised too

test/test_sequence.py
```diff
@@ -0,0 +1,84 @@
...
+    autoincrement_table.create(engine_testaccount)
+    try:
+        engine_testaccount.execute(autoincrement_table.insert(), [{'data': 'test_insert_1'}])
+
+        select_stmt = select([autoincrement_table])
+        result = engine_testaccount.execute(select_stmt).fetchall()
+        assert result == [(1, 'test_insert_1')]
+
+        metadata_autoload = MetaData()
+        autoload_sequence_table = Table(test_table_name, metadata_autoload, autoload=True, autoload_with=engine_testaccount)
```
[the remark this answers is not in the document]
sfc-gh-aling: updated

test/test_sequence.py
sfc-gh-aling: that's a good piece of advice, updated.

test/test_sequence.py
sfc-gh-aling: removed

test/test_sequence.py
sfc-gh-aling: I see other old tests are returning, code copied from them.
removing the returning doesn't impact anything, I will simply remove it

test/test_sequence.py
```diff
@@ -0,0 +1,84 @@
...
+
+        engine_testaccount.execute(autoload_sequence_table.insert(),
+                                   [{'data': 'multi_insert_1'}, {'data': 'multi_insert_2'}])
+
+        engine_testaccount.execute(autoload_sequence_table.insert(), [{'data': 'test_insert_2'}])
+
+        result = engine_testaccount.execute(select_stmt).fetchall()
```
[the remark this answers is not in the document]
sfc-gh-aling: updated

test/test_sequence.py
```diff
@@ -0,0 +1,84 @@
...
+
+from sqlalchemy import Column, Integer, MetaData, Sequence, String, Table, select
+
+
+def test_table_with_sequence(engine_testaccount, db_parameters):
+    # https://github.com/snowflakedb/snowflake-sqlalchemy/issues/124
+    metadata = MetaData()
+    test_table_name = 'sequence'
+    sequence_table = Table(test_table_name, metadata,
+                           Column('id', Integer, Sequence(test_table_name + '_id_seq'), primary_key=True),
```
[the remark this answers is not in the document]
sfc-gh-aling: good catch, I updated code to stay consistent

test/test_sequence.py
sfc-gh-aling: other tests might also want to be updated to not return, but I think we could make the change incrementally when we touch those tests.

sfc-gh-mkeller (approved):
Ship it 🚢 🚢

PR #331, opened 2022-08-02
SNOW-633094: Switch all source code to use f-strings and with statements

sfc-gh-aling:

   Fixes SNOW-633094

Improve code quality including using f string, formatting tests.

### DIFF
--- a/src/snowflake/sqlalchemy/custom_commands.py
+++ b/src/snowflake/sqlalchemy/custom_commands.py
@@ -42,12 +42,7 @@ class MergeInto(UpdateBase):
         def __repr__(self):
             if self.command == "INSERT":
                 sets, sets_tos = zip(*self.set.items())
-                return "WHEN NOT MATCHED{} THEN {} ({}) VALUES ({})".format(
-                    " AND %s" % self.predicate if self.predicate is not None else "",
-                    self.command,
-                    ", ".join(sets),
-                    ", ".join(map(str, sets_tos)),
-                )
+                return f"""WHEN NOT MATCHED{f' AND {str(self.predicate if self.predicate is not None else "")}'} THEN {self.command} ({", ".join(sets)}) VALUES ({", ".join(map(str, sets_tos))})"""
             else:
                 # WHEN MATCHED clause
                 sets = (
@@ -55,11 +50,7 @@ class MergeInto(UpdateBase):
                     if self.set
                     else ""
                 )
-                return "WHEN MATCHED{} THEN {}{}".format(
-                    " AND %s" % self.predicate if self.predicate is not None else "",
-                    self.command,
-                    " SET %s" % sets if self.set else "",
-                )
+                return f"""WHEN MATCHED{f' AND {str(self.predicate if self.predicate is not None else "")}'} THEN {self.command}{f' SET {str(sets if self.set else "")}'}"""
 
         def values(self, **kwargs):
             self.set = kwargs
@@ -509,21 +500,14 @@ class AWSBucket(ClauseElement):
         return cls(bucket, path)
 
     def __repr__(self):
-        credentials = "CREDENTIALS=({})".format(
-            " ".join(f"{n}='{v}'" for n, v in self.credentials_used.items())
+        credentials = f"""CREDENTIALS=({" ".join(f"{n}='{v}'" for n, v in self.credentials_used.items())})"""
+        encryption_value = " ".join(
+            (f"{n}='{v}'" if isinstance(v, string_types) else "{n}={v}")
+            for n, v in self.encryption_used.items()
         )
-        encryption = "ENCRYPTION=({})".format(
-            " ".join(
-                ("{}='{}'" if isinstance(v, string_types) else "{}={}").format(n, v)
-                for n, v in self.encryption_used.items()
-            )
-        )
-        uri = "'s3://{}{}'".format(self.bucket, "/" + self.path if self.path else "")
-        return "{}{}{}".format(
-            uri,
-            " " + credentials if self.credentials_used else "",
-            " " + encryption if self.encryption_used else "",
-        )
+        encryption = f"ENCRYPTION=({encryption_value})"
+        uri = f"""'s3://{self.bucket}{"/" + self.path if self.path else ""}'"""
+        return f'{uri}{" " + credentials if self.credentials_used else ""}{" " + encryption if self.encryption_used else ""}'
 
     def credentials(
         self, aws_role=None, aws_key_id=None, aws_secret_key=None, aws_token=None
@@ -585,23 +569,14 @@ class AzureContainer(ClauseElement):
         return cls(account, container, path)
 
     def __repr__(self):
-        credentials = "CREDENTIALS=({})".format(
-            " ".join(f"{n}='{v}'" for n, v in self.credentials_used.items())
+        credentials = f"""CREDENTIALS=({" ".join(f"{n}='{v}'" for n, v in self.credentials_used.items())})"""
+        encryption_value = " ".join(
+            (f"{n}='{v}'" if isinstance(v, string_types) else "{n}={v}")
+            for n, v in self.encryption_used.items()
         )
-        encryption = "ENCRYPTION=({})".format(
-            " ".join(
-                ("{}='{}'" if isinstance(v, string_types) else "{}={}").format(n, v)
-                for n, v in self.encryption_used.items()
-            )
-        )
-        uri = "'azure://{}.blob.core.windows.net/{}{}'".format(
-            self.account, self.container, "/" + self.path if self.path else ""
-        )
-        return "{}{}{}".format(
-            uri,
-            f" {credentials}" if self.credentials_used else "",
-            f" {encryption}" if self.encryption_used else "",
-        )
+        encryption = f"ENCRYPTION=({encryption_value})"
+        uri = f"""'azure://{self.account}.blob.core.windows.net/{self.container}{"/" + self.path if self.path else ""}'"""
+        return f'{uri}{f" {credentials}" if self.credentials_used else ""}{f" {encryption}" if self.encryption_used else ""}'
 
     def credentials(self, azure_sas_token):
         self.credentials_used = {"AZURE_SAS_TOKEN": azure_sas_token}
--- a/src/snowflake/sqlalchemy/custom_commands.py
+++ b/src/snowflake/sqlalchemy/custom_commands.py
@@ -46,7 +46,7 @@ class MergeInto(UpdateBase):
 WHEN NOT MATCHED\
 {f" AND {str(self.predicate if self.predicate is not None else '')}"} \
 THEN {self.command} ({', '.join(sets)}) \
-VALUES ({', '.join(map(str, sets_tos))})
+VALUES ({', '.join(map(str, sets_tos))})\
 """
             else:
                 # WHEN MATCHED clause
@@ -59,7 +59,7 @@ VALUES ({', '.join(map(str, sets_tos))})
 WHEN MATCHED\
 {f" AND {str(self.predicate if self.predicate is not None else '')}"} \
 THEN {self.command}\
-{f"SET {str(sets if self.set else '')}"}
+{f"SET {str(sets if self.set else '')}"}\
 """
 
         def values(self, **kwargs):
@@ -512,12 +512,12 @@ class AWSBucket(ClauseElement):
     def __repr__(self):
         credentials = f"""CREDENTIALS=({" ".join(f"{n}='{v}'" for n, v in self.credentials_used.items())})"""
         encryption_value = " ".join(
-            (f"{n}='{v}'" if isinstance(v, string_types) else "{n}={v}")
+            (f"{n}='{v}'" if isinstance(v, string_types) else f"{n}={v}")
             for n, v in self.encryption_used.items()
         )
         encryption = f"ENCRYPTION=({encryption_value})"
         uri = f"""'s3://{self.bucket}{"/" + self.path if self.path else ""}'"""
-        return f"{uri}{' ' + credentials if self.credentials_used else ''}{' ' + encryption if self.encryption_used else ''}"
+        return f"{uri}{' ' + credentials if self.credentials_used else ''}{f' {encryption}' if self.encryption_used else ''}"
 
     def credentials(
         self, aws_role=None, aws_key_id=None, aws_secret_key=None, aws_token=None
@@ -581,11 +581,11 @@ class AzureContainer(ClauseElement):
     def __repr__(self):
         credentials = f"""CREDENTIALS=({" ".join(f"{n}='{v}'" for n, v in self.credentials_used.items())})"""
         encryption_value = " ".join(
-            (f"{n}='{v}'" if isinstance(v, string_types) else "{n}={v}")
+            (f"{n}='{v}'" if isinstance(v, string_types) else f"{n}={v}")
             for n, v in self.encryption_used.items()
         )
         encryption = f"ENCRYPTION=({encryption_value})"
-        uri = f"""'azure://{self.account}.blob.core.windows.net/{self.container}{"/" + self.path if self.path else ""}'"""
+        uri = f"""'azure://{self.account}.blob.core.windows.net/{self.container}{f"/{self.path}" if self.path else ""}'"""
         return f"{uri}{f' {credentials}' if self.credentials_used else ''}{f' {encryption}' if self.encryption_used else ''}"
 
     def credentials(self, azure_sas_token):
--- a/src/snowflake/sqlalchemy/custom_commands.py
+++ b/src/snowflake/sqlalchemy/custom_commands.py
@@ -40,14 +40,17 @@ class MergeInto(UpdateBase):
             self.command = command
 
         def __repr__(self):
+            case_predicate = (
+                f" AND {str(self.predicate)}" if self.predicate is not None else ""
+            )
             if self.command == "INSERT":
                 sets, sets_tos = zip(*self.set.items())
-                return f"""\
-WHEN NOT MATCHED\
-{f" AND {str(self.predicate) if self.predicate is not None else ''}"} \
-THEN {self.command} ({', '.join(sets)}) \
-VALUES ({', '.join(map(str, sets_tos))})\
-"""
+                return "WHEN NOT MATCHED{} THEN {} ({}) VALUES ({})".format(
+                    case_predicate,
+                    self.command,
+                    ", ".join(sets),
+                    ", ".join(map(str, sets_tos)),
+                )
             else:
                 # WHEN MATCHED clause
                 sets = (
@@ -55,12 +58,11 @@ VALUES ({', '.join(map(str, sets_tos))})\
                     if self.set
                     else ""
                 )
-                return f"""\
-WHEN MATCHED\
-{f" AND {str(self.predicate if self.predicate is not None else '')}"} \
-THEN {self.command}\
-{f"SET {str(sets if self.set else '')}"}\
-"""
+                return "WHEN MATCHED{} THEN {}{}".format(
+                    case_predicate,
+                    self.command,
+                    f" SET {str(sets)}" if self.set else "",
+                )
 
         def values(self, **kwargs):
             self.set = kwargs
@@ -73,7 +75,7 @@ THEN {self.command}\
     def __repr__(self):
         clauses = " ".join([repr(clause) for clause in self.clauses])
         return f"MERGE INTO {self.target} USING {self.source} ON {self.on}" + (
-            " " + clauses if clauses else ""
+            f" {clauses}" if clauses else ""
         )
 
     def when_matched_then_update(self):
@@ -451,7 +453,7 @@ class ExternalStage(ClauseElement, FromClauseRole):
         """
         return cls(
             parent_stage.name,
-            parent_stage.path + "/" + path,
+            f"{parent_stage.path}/{path}",
             parent_stage.namespace,
             file_format,
         )
@@ -510,14 +512,21 @@ class AWSBucket(ClauseElement):
         return cls(bucket, path)
 
     def __repr__(self):
-        credentials = f"""CREDENTIALS=({" ".join(f"{n}='{v}'" for n, v in self.credentials_used.items())})"""
-        encryption_value = " ".join(
-            (f"{n}='{v}'" if isinstance(v, string_types) else f"{n}={v}")
-            for n, v in self.encryption_used.items()
+        credentials = "CREDENTIALS=({})".format(
+            " ".join(f"{n}='{v}'" for n, v in self.credentials_used.items())
         )
-        encryption = f"ENCRYPTION=({encryption_value})"
-        uri = f"""'s3://{self.bucket}{"/" + self.path if self.path else ""}'"""
-        return f"{uri}{' ' + credentials if self.credentials_used else ''}{f' {encryption}' if self.encryption_used else ''}"
+        encryption = "ENCRYPTION=({})".format(
+            " ".join(
+                f"{n}='{v}'" if isinstance(v, string_types) else f"{n}={v}"
+                for n, v in self.encryption_used.items()
+            )
+        )
+        uri = "'s3://{}{}'".format(self.bucket, f"/{self.path}" if self.path else "")
+        return "{}{}{}".format(
+            uri,
+            f" {credentials}" if self.credentials_used else "",
+            f" {encryption}" if self.encryption_used else "",
+        )
 
     def credentials(
         self, aws_role=None, aws_key_id=None, aws_secret_key=None, aws_token=None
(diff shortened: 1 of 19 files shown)


src/snowflake/sqlalchemy/custom_commands.py
```diff
@@ -42,24 +42,25 @@ def __init__(self, command):
         def __repr__(self):
             if self.command == "INSERT":
                 sets, sets_tos = zip(*self.set.items())
-                return "WHEN NOT MATCHED{} THEN {} ({}) VALUES ({})".format(
-                    " AND %s" % self.predicate if self.predicate is not None else "",
-                    self.command,
-                    ", ".join(sets),
-                    ", ".join(map(str, sets_tos)),
-                )
+                return f"""
```
sfc-gh-mkeller, line 45:
```suggestion
                return f"""\
```

src/snowflake/sqlalchemy/custom_commands.py
```diff
@@ -42,24 +42,25 @@ def __init__(self, command):
...
+WHEN NOT MATCHED\
+{f" AND {str(self.predicate if self.predicate is not None else '')}"} \
```
sfc-gh-mkeller, line 47: Isn't `str` redundant here? I always thought that Python will call `str` on f-string pieces because it's embedding it in a string.

```pycon
>>> class A:
...     def __init__(self, val=None):
...             self.val = val
...     def __str__(self):
...             print("str")
...             return str(self.val)
...
>>> a=A('a')
>>> print(f"{a}")
str
a
>>> b=A('b')
>>> print(f"{b}")
str
b
```

src/snowflake/sqlalchemy/custom_commands.py
```diff
@@ -42,24 +42,25 @@ def __init__(self, command):
...
+THEN {self.command} ({', '.join(sets)}) \
+VALUES ({', '.join(map(str, sets_tos))})
```
sfc-gh-mkeller, line 49:
```suggestion
VALUES ({', '.join(map(str, sets_tos))})\
```

src/snowflake/sqlalchemy/custom_commands.py
```diff
@@ -42,24 +42,25 @@ def __init__(self, command):
...
                     else ""
                 )
-                return "WHEN MATCHED{} THEN {}{}".format(
-                    " AND %s" % self.predicate if self.predicate is not None else "",
-                    self.command,
-                    " SET %s" % sets if self.set else "",
-                )
+                return f"""\
+WHEN MATCHED\
+{f" AND {str(self.predicate if self.predicate is not None else '')}"} \
```
sfc-gh-mkeller, line 60: Same for `str` here

src/snowflake/sqlalchemy/custom_commands.py
```diff
@@ -42,24 +42,25 @@ def __init__(self, command):
...
+THEN {self.command}\
+{f"SET {str(sets if self.set else '')}"}
```
sfc-gh-mkeller, line 62:
```suggestion
{f"SET {str(sets if self.set else '')}"}\
```

src/snowflake/sqlalchemy/custom_commands.py
```diff
@@ -509,21 +510,14 @@ def from_uri(cls, uri):
...
-            )
-        )
-        uri = "'s3://{}{}'".format(self.bucket, "/" + self.path if self.path else "")
-        return "{}{}{}".format(
-            uri,
-            " " + credentials if self.credentials_used else "",
-            " " + encryption if self.encryption_used else "",
+        credentials = f"""CREDENTIALS=({" ".join(f"{n}='{v}'" for n, v in self.credentials_used.items())})"""
+        encryption_value = " ".join(
+            (f"{n}='{v}'" if isinstance(v, string_types) else "{n}={v}")
```
sfc-gh-mkeller, line 515:
```suggestion
            (f"{n}='{v}'" if isinstance(v, string_types) else f"{n}={v}")
```

src/snowflake/sqlalchemy/custom_commands.py
```diff
@@ -509,21 +510,14 @@ def from_uri(cls, uri):
...
+            for n, v in self.encryption_used.items()
         )
+        encryption = f"ENCRYPTION=({encryption_value})"
+        uri = f"""'s3://{self.bucket}{"/" + self.path if self.path else ""}'"""
+        return f"{uri}{' ' + credentials if self.credentials_used else ''}{' ' + encryption if self.encryption_used else ''}"
```
sfc-gh-mkeller, line 520:
```suggestion
        return f"{uri}{' ' + credentials if self.credentials_used else ''}{f' {encryption}' if self.encryption_used else ''}"
```

src/snowflake/sqlalchemy/custom_commands.py
```diff
@@ -585,23 +579,14 @@ def from_uri(cls, uri):
...
-        uri = "'azure://{}.blob.core.windows.net/{}{}'".format(
-            self.account, self.container, "/" + self.path if self.path else ""
-        )
-        return "{}{}{}".format(
-            uri,
-            f" {credentials}" if self.credentials_used else "",
-            f" {encryption}" if self.encryption_used else "",
+        credentials = f"""CREDENTIALS=({" ".join(f"{n}='{v}'" for n, v in self.credentials_used.items())})"""
+        encryption_value = " ".join(
+            (f"{n}='{v}'" if isinstance(v, string_types) else "{n}={v}")
```
sfc-gh-mkeller, line 584:
```suggestion
            (f"{n}='{v}'" if isinstance(v, string_types) else f"{n}={v}")
```

src/snowflake/sqlalchemy/custom_commands.py
```diff
@@ -585,23 +579,14 @@ def from_uri(cls, uri):
...
+            for n, v in self.encryption_used.items()
         )
+        encryption = f"ENCRYPTION=({encryption_value})"
+        uri = f"""'azure://{self.account}.blob.core.windows.net/{self.container}{"/" + self.path if self.path else ""}'"""
```
sfc-gh-mkeller, line 588:
```suggestion
        uri = f"""'azure://{self.account}.blob.core.windows.net/{self.container}{f"/{self.path}" if self.path else ""}'"""
```

src/snowflake/sqlalchemy/util.py
```diff
@@ -20,27 +20,13 @@ def _url(**db_parameters):
         raise exc.ArgumentError("account parameter must be specified.")
 
     if "host" in db_parameters:
-        ret = "snowflake://{user}:{password}@{host}:{port}/".format(
-            user=db_parameters.get("user", ""),
-            password=_rfc_1738_quote(db_parameters.get("password", "")),
-            host=db_parameters["host"],
-            port=db_parameters["port"] if "port" in db_parameters else 443,
-        )
+        ret = f"snowflake://{db_parameters.get('user', '')}:{_rfc_1738_quote(db_parameters.get('password', ''))}@{db_parameters['host']}:{db_parameters['port'] if 'port' in db_parameters else 443}/"
```
sfc-gh-mkeller, line 23:
```suggestion
        ret = f"snowflake://{db_parameters.get('user', '')}:{_rfc_1738_quote(db_parameters.get('password', ''))}@{db_parameters['host']}:{db_parameters.get('port', 443)}/"
```

src/snowflake/sqlalchemy/util.py
```diff
@@ -20,27 +20,13 @@ def _url(**db_parameters):
...
+        ret = f"snowflake://{db_parameters.get('user', '')}:{_rfc_1738_quote(db_parameters.get('password', ''))}@{db_parameters['account']}/"
         specified_parameters += ["user", "password", "account"]
     else:
-        ret = "snowflake://{user}:{password}@{account}.{region}/".format(
-            account=db_parameters["account"],
-            user=db_parameters.get("user", ""),
-            password=_rfc_1738_quote(db_parameters.get("password", "")),
-            region=db_parameters["region"],
-        )
+        ret = f"snowflake://{db_parameters.get('user', '')}:{_rfc_1738_quote(db_parameters.get('password', ''))}@{db_parameters['account']}.{db_parameters['region']}/"
```
sfc-gh-mkeller, line 29: We could add port to this too

src/snowflake/sqlalchemy/util.py
```diff
@@ -20,27 +20,13 @@ def _url(**db_parameters):
...
-        )
+        ret = f"snowflake://{db_parameters.get('user', '')}:{_rfc_1738_quote(db_parameters.get('password', ''))}@{db_parameters['host']}:{db_parameters['port'] if 'port' in db_parameters else 443}/"
         specified_parameters += ["user", "password", "host", "port"]
     elif "region" not in db_parameters:
-        ret = "snowflake://{user}:{password}@{account}/".format(
-            account=db_parameters["account"],
-            user=db_parameters.get("user", ""),
-            password=_rfc_1738_quote(db_parameters.get("password", "")),
-        )
+        ret = f"snowflake://{db_parameters.get('user', '')}:{_rfc_1738_quote(db_parameters.get('password', ''))}@{db_parameters['account']}/"
```
sfc-gh-mkeller, line 26: This is missing port

tests/conftest.py
```diff
@@ -24,7 +24,7 @@
 RUNNING_ON_GH = os.getenv("GITHUB_ACTIONS") == "true"
 
 if os.getenv("TRAVIS") == "true":
-    TEST_SCHEMA = "TRAVIS_JOB_{}".format(os.getenv("TRAVIS_JOB_ID"))
+    TEST_SCHEMA = f"TRAVIS_JOB_{os.getenv('TRAVIS_JOB_ID')}"
```
sfc-gh-mkeller, line 27: We don't use TravisCI anymore, you can get rid of this

tests/test_copy.py
```diff
@@ -155,25 +154,25 @@ def test_copy_into_location(engine_testaccount, sql_compiler):
...
-            ):
-                raise Exception(
-                    f"Not acceptable exception: {str(exc)} {str(exc.value)}"
-                )
+        with engine_testaccount.connect() as conn:
+            acceptable_exc_reasons = {
+                "Failure using stage area",
+                "AWS_ROLE credentials are not allowed for this account.",
+                "AWS_ROLE credentials are invalid",
+            }
```
sfc-gh-mkeller, line 162: I personally like to define constants before making a connection, but this is just a nit pick and not super important

tests/test_geography.py
```diff
@@ -44,26 +44,27 @@ def test_inspect_geography_datatypes(engine_testaccount):
...
 
-        # select
-        s = select(test_geography)
-        results = conn.execute(s)
-        rows = results.fetchone()
-        results.close()
-        assert rows[0] == 1
-        assert rows[1] == rows[2]
-        assert loads(rows[2]) == loads(test_point1)
+            # select
```
sfc-gh-mkeller, line 61:
```suggestion
```

tests/test_orm.py
```diff
@@ -250,9 +250,7 @@ def test_schema_including_dot(engine_testaccount, db_parameters, run_v20_sqlalch
     """
     Base = declarative_base()
 
-    namespace = '{db}."{schema}.{schema}".{db}'.format(
-        db=db_parameters["database"].lower(), schema=db_parameters["schema"].lower()
-    )
+    namespace = f'{db_parameters["database"].lower()}."{db_parameters["schema"].lower()}.{db_parameters["schema"].lower()}".{db_parameters["database"].lower()}'
```
sfc-gh-mkeller, line 253: So in these cases I'd rather have `format` calls (that is when we repeat arguments).
Now this is in the tests, so it doesn't matter too much, but also performance isn't that big of a priority in (unlike in the actual library).

tests/test_orm.py
```diff
@@ -268,9 +266,7 @@ class User(Base):
     session.future = run_v20_sqlalchemy
     query = session.query(User.id)
     assert str(query).startswith(
-        'SELECT {db}."{schema}.{schema}".{db}.users.id'.format(
-            db=db_parameters["database"].lower(), schema=db_parameters["schema"].lower()
-        )
+        f'SELECT {db_parameters["database"].lower()}."{db_parameters["schema"].lower()}.{db_parameters["schema"].lower()}".{db_parameters["database"].lower()}.users.id'
```
sfc-gh-mkeller, line 269: Same as above, I think `format` is more appropriate here.

tests/test_timestamp.py
```diff
@@ -66,19 +66,19 @@ def test_inspect_timestamp_datatypes(engine_testaccount):
...
-        # select
-        s = select(test_timestamp)
-        results = conn.execute(s)
-        rows = results.fetchone()
-        results.close()
-        assert rows[0] == 1
-        assert rows[1] == current_utctime
-        assert rows[2] == current_localtime
-        assert rows[3] == current_localtime_with_other_tz
+            # select
```
sfc-gh-mkeller, line 74:
```suggestion
```

tests/test_sequence.py
```diff
@@ -17,46 +17,47 @@ def test_table_with_sequence(engine_testaccount, db_parameters):
...
-        with conn.begin():
-            result = conn.execute(select_stmt).fetchall()
-            assert result == [(1, "test_insert_1")]
-
-        autoload_sequence_table = Table(
-            test_table_name, MetaData(), autoload_with=engine_testaccount
-        )
+            with conn.begin():
+                result = conn.execute(select_stmt).fetchall()
+                assert result == [(1, "test_insert_1")]
```
sfc-gh-mkeller, line 29: I think it'd also be a good idea to standardize result fetching to either always happen in, or out of the `begin` block. Just in case users take inspiration from our tests

tests/test_pandas.py
```diff
@@ -74,36 +75,37 @@ def test_a_simple_read_sql(engine_testaccount):
...
+            with conn.begin():
+                results = conn.execute(ins)
+            # Note: SQLAlchemy 1.4 changed what ``inserted_primary_key`` returns
+            #  a cast is here to make sure the test works with both older and newer
+            #  versions
+            assert list(results.inserted_primary_key) == [1], "sequence value"
+            results.close()
+
+            # inserts data with the given id
+            ins = users.insert()
```
sfc-gh-mkeller, line 91: This is only used once, let's inline it

tests/test_pandas.py
```diff
@@ -142,71 +144,62 @@ def get_engine_with_numpy(db_parameters, user=None, password=None, account=None)
...
-    conn.exec_driver_sql(
+    with engine.connect() as conn:
+        total_rows = 10000
+        conn.exec_driver_sql(
+            textwrap.dedent(
+                f"""\
+            create or replace table src(c1 float)
+            as select random(123) from table(generator(timelimit=>1))
+            limit {total_rows}
         """
```
sfc-gh-mkeller, line 179: If the indent level doesn't match then `dedent` will not remove all spaces here
```suggestion
            """
```

tests/test_pandas.py
```diff
@@ -242,65 +235,68 @@ def test_timezone(db_parameters):
...
+            conn.exec_driver_sql(
+                textwrap.dedent(
+                    f"""\
+                CREATE OR REPLACE TABLE {test_table_name}(
+                    tz_col timestamp_tz,
+                    ntz_col timestamp_ntz,
+                    decimal_col decimal(10,1),
+                    float_col float
+                );
             """
```
sfc-gh-mkeller, line 250:
```suggestion
                """
```

tests/test_pandas.py
```diff
@@ -242,65 +235,68 @@ def test_timezone(db_parameters):
...
+                conn.exec_driver_sql(
+                    textwrap.dedent(
+                        f"""\
+                    INSERT INTO {test_table_name}
+                        SELECT
+                            current_timestamp(),
+                            current_timestamp()::timestamp_ntz,
+                            to_decimal(.1, 10, 1),
+                            .10;
                 """
```
sfc-gh-mkeller, line 265:
```suggestion
                    """
```

tests/test_pandas.py
```diff
@@ -353,22 +345,23 @@ def write_to_db():
...
-            )
-            # Verify that all 10 entries were written to the DB
-            for i in range(10):
-                assert results[i] == (i, i + 10)
-            assert len(results) == 10
-    finally:
-        conn.exec_driver_sql(f"DROP TABLE IF EXISTS {table_name}")
+                results = sorted(
+                    conn.exec_driver_sql(f"SELECT * FROM {table_name}").fetchall(),
+                    key=lambda x: x[0],
```
sfc-gh-mkeller, line 360: You could use [itemgetter](https://docs.python.org/3/library/operator.html#operator.itemgetter) here

tests/test_core.py
```diff
@@ -102,17 +103,8 @@ def test_connect_args():
...
-            user=CONNECTION_PARAMETERS["user"],
-            account=CONNECTION_PARAMETERS["account"],
-            password=CONNECTION_PARAMETERS["password"],
-            host=CONNECTION_PARAMETERS["host"],
-            port=CONNECTION_PARAMETERS["port"],
-            database=CONNECTION_PARAMETERS["database"],
-            schema=CONNECTION_PARAMETERS["schema"],
-            protocol=CONNECTION_PARAMETERS["protocol"],
-        )
+        f"""snowflake://{CONNECTION_PARAMETERS["user"]}:{CONNECTION_PARAMETERS["password"]}@{CONNECTION_PARAMETERS["host"]}:{CONNECTION_PARAMETERS["port"]}/{CONNECTION_PARAMETERS["database"]}/{CONNECTION_PARAMETERS["schema"]}"
```
sfc-gh-mkeller, line 106: This line is too long

tests/test_core.py
```diff
@@ -102,17 +103,8 @@ def test_connect_args():
...
+        "?account={CONNECTION_PARAMETERS["account"]}&protocol={CONNECTION_PARAMETERS["protocol"]}"""
```
sfc-gh-mkeller, line 107: I believe that this was a mistake
```suggestion
        f"""snowflake://{CONNECTION_PARAMETERS["user"]}:{CONNECTION_PARAMETERS["password"]}@{CONNECTION_PARAMETERS["host"]}:{CONNECTION_PARAMETERS["port"]}/{CONNECTION_PARAMETERS["database"]}/{CONNECTION_PARAMETERS["schema"]}\
        ?account={CONNECTION_PARAMETERS["account"]}&protocol={CONNECTION_PARAMETERS["protocol"]}"""
```

tests/test_core.py
```diff
@@ -192,179 +184,192 @@ def test_insert_tables(engine_testaccount):
...
+            #  a cast is here to make sure the test works with both older and newer
+            #  versions
+            assert list(results.inserted_primary_key) == [1], "sequence value"
+            results.close()
 
-        # verify the results
-        with conn.begin():
-            s = select(users)
+            # inserts data with the given id
+            ins = users.insert()
```
sfc-gh-mkeller, line 200: This can be inlined

tests/test_core.py
```diff
@@ -192,179 +184,192 @@ def test_insert_tables(engine_testaccount):
...
+                conn.execute(
+                    ins, {"id": 2, "name": "wendy", "fullname": "Wendy Williams"}
+                )
+
+            # verify the results
+            with conn.begin():
+                s = select(users)
+                results = conn.execute(s)
+                assert (
+                    len([row for row in results]) == 2
```
sfc-gh-mkeller, line 211:
```suggestion
                    len(results) == 2
```

tests/test_core.py
```diff
@@ -736,26 +731,23 @@ def test_get_temp_table_names(engine_testaccount):
...
 
 
 def test_create_table_with_schema(engine_testaccount, db_parameters):
     metadata = MetaData()
-    new_schema = db_parameters["schema"] + "_NEW"
+    new_schema = (
+        db_parameters["schema"]
+        + "_NEW_"
+        + random_string(5, choices=string.ascii_uppercase)
+    )
```
sfc-gh-mkeller, line 750:
```suggestion
    new_schema = f"{db_parameters["schema"]}_NEW_{random_string(5, choices=string.ascii_uppercase)}"
```
Not sure if this fits in one line

sfc-gh-mkeller (changes requested):

src/snowflake/sqlalchemy/custom_commands.py
sfc-gh-aling: I called `str` because "%s" was used in the original code. I did some more search and it's another rabbit hole: https://peps.python.org/pep-0498/#format-specifiers

but in short, the call priority chain is `__format__` -> `__str__` -> `__repr__` when using f-strings in my test.
so I think `str` is needed here just being cautious.

```python
class A:
    def __str__(self):
        return 'str called'

    def __repr__(self):
        return 'repr called'

    def __format__(self, format_spec):
        return 'format called'


a = A()
str_a = f"{a}"
print(str_a)  # 'format called'
```

src/snowflake/sqlalchemy/util.py
sfc-gh-aling: may I ask to which part we could add the port? is it after the region?
I checked the samples looks like all port come with host

src/snowflake/sqlalchemy/custom_commands.py
sfc-gh-aling: same comment as above, `str()` cast for equivalent "%s"

src/snowflake/sqlalchemy/util.py
sfc-gh-aling: okay, I tried to add it but found it might become a behavior breaking change for the `URL` method (or  a bug fix depending on how we look at it).

- previously if we do `URL(account="testaccount", user="admin", password="test", database="testdb", port=8080)`, we get string `snowflake://admin:test@testaccount/testdb?port=8080` back.
- if we made the change, it would be `snowflake://admin:test@testaccount:8080/testdb`

both style could get the `port` populated correctly, but the string itself changes.. so I'm a bit reluctant to introduce the change unless users ask for it.

thoughts?

tests/test_sequence.py
sfc-gh-aling: good suggestion, I updated all the tests to have assertion be within `begin()`

src/snowflake/sqlalchemy/util.py
sfc-gh-aling: see comment above

tests/test_core.py
```diff
@@ -199,182 +199,196 @@ def test_insert_tables(engine_testaccount):
...
-                    {"user_id": 1, "email_address": "jack@yahoo.com"},
-                    {"user_id": 1, "email_address": "jack@msn.com"},
-                    {"user_id": 2, "email_address": "www@www.org"},
-                    {"user_id": 2, "email_address": "wendy@aol.com"},
-                ],
-            )
+                # verify the results
+                results = conn.execute(select(users))
+                assert (
+                    len([row for row in results]) == 2
```
sfc-gh-aling, line 224:
```suggestion
                    len(results) == 2
```

tests/test_pandas.py
```diff
@@ -142,71 +145,60 @@ def get_engine_with_numpy(db_parameters, user=None, password=None, account=None)
...
-            total_rows
+    with engine.connect() as conn:
+        total_rows = 10000
+        conn.exec_driver_sql(
+            textwrap.dedent(
+                f"""\
+            create or replace table src(c1 float)
+            as select random(123) from table(generator(timelimit=>1))
+            limit {total_rows}
+            """
```
sfc-gh-mkeller, line 178: Looks a little weird
```suggestion
                f"""\
                create or replace table src(c1 float)
                as select random(123) from table(generator(timelimit=>1))
                limit {total_rows}
                """
```

tests/test_qmark.py
```diff
@@ -55,32 +55,26 @@ def test_qmark_bulk_insert(db_parameters, run_v20_sqlalchemy):
...
+
+                for data in pd.read_sql_query(
+                    text("select * from src"), con, chunksize=16000
+                ):
+                    data.to_sql(
+                        "dst", con, if_exists="append", index=False, index_label=None
+                    )
+
+        finally:
+            engine.dispose()
```
sfc-gh-mkeller, line 79: We're inside of `engine`'s context manager, is `dispose` necessary here?

tests/test_semi_structured_datatypes.py
```diff
@@ -74,36 +75,37 @@ def test_inspect_semi_structured_datatypes(engine_testaccount):
...
+                    f"""
+                    INSERT INTO {table_name}(id, va, ar)
+                    SELECT 1,
+                           PARSE_JSON('{{"vk1":100, "vk2":200, "vk3":300}}'),
+                           PARSE_JSON('[
+                    {{"k":1, "v":"str1"}},
+                    {{"k":2, "v":"str2"}},
+                    {{"k":3, "v":"str3"}}]'
+                    )
+                """
```
sfc-gh-mkeller, line 91:
```suggestion
                    """
```

tests/test_core.py
```diff
@@ -199,182 +199,194 @@ def test_insert_tables(engine_testaccount):
...
+                    users.c.id == addresses.c.user_id,
+                    or_(
+                        addresses.c.email_address == "wendy@aol.com",
+                        addresses.c.email_address == "jack@yahoo.com",
+                    ),
+                    not_(users.c.id > 5),
+                )
+                expected_sql = textwrap.dedent(
+                    """\
+                users.name LIKE :name_1
```
sfc-gh-mkeller, line 301:
```suggestion
                 users.name LIKE :name_1
```

tests/test_core.py
```diff
@@ -1372,32 +1342,30 @@ def test_special_schema_character(db_parameters, on_public_ci):
...
+                "select current_database(), " "current_schema();"
+            )
+        ]
+    with create_engine(URL(**options)).connect() as sa_conn:
+        sa_connection = [
+            res
+            for res in sa_conn.execute(
+                text("select current_database(), " "current_schema();")
+            )
+        ]
```
sfc-gh-mkeller, line 1364:
```suggestion
        sa_connection = sa_conn.execute(
                text("select current_database(), " "current_schema();")
        ).fetchall()
```

tests/test_core.py
```diff
@@ -1372,32 +1342,30 @@ def test_special_schema_character(db_parameters, on_public_ci):
...
-    ]
-    sa_conn.close()
-    sf_conn.close()
+    with connect(**options) as sf_conn:
+        sf_connection = [
+            res
+            for res in sf_conn.cursor().execute(
+                "select current_database(), " "current_schema();"
+            )
+        ]
```
sfc-gh-mkeller, line 1357:
```suggestion
        sf_connection = sf_conn.cursor().execute(
                "select current_database(), " "current_schema();"
        ).fetchall()
```

sfc-gh-mkeller (approved):
Other than small comments

:ship: :ship:

tests/test_core.py
sfc-gh-aling: actually we need the extra space such that precommit won't strip the space for us..

tests/test_qmark.py
sfc-gh-aling: `dispose` will shutdown the connection pool, so I feel we could keep it here showing the usage pattern to users

PR #375, opened 2023-01-26
SNOW-733235: Returning correct exception when table does not exist

davidov541:
…ption.

   Fixes SNOW-733235

   My code detects if a table does not exist in a dictionary before attempting to index into that dictionary during requests to get columns. This prevents KeyErrors from being returned, and allows us to instead return more descriptive and standard errors indicating that the table does not exist.

### DIFF
--- a/src/snowflake/sqlalchemy/snowdialect.py
+++ b/src/snowflake/sqlalchemy/snowdialect.py
@@ -637,6 +637,9 @@ class SnowflakeDialect(default.DefaultDialect):
 
             type_instance = col_type(**col_type_kw)
 
+            if table_name not in schema_primary_keys:
+                raise sa_exc.NoSuchTableError()
+
             current_table_pks = schema_primary_keys.get(table_name)
 
             ans.append(
@@ -669,7 +672,10 @@ class SnowflakeDialect(default.DefaultDialect):
         if schema_columns is None:
             # Too many results, fall back to only query about single table
             return self._get_table_columns(connection, table_name, schema, **kw)
-        return schema_columns[self.normalize_name(table_name)]
+        normalized_table_name = self.normalize_name(table_name)
+        if normalized_table_name not in schema_columns:
+            raise sa_exc.NoSuchTableError()
+        return schema_columns[normalized_table_name]
 
     @reflection.cache
     def get_table_names(self, connection, schema=None, **kw):
--- a/src/snowflake/sqlalchemy/snowdialect.py
+++ b/src/snowflake/sqlalchemy/snowdialect.py
@@ -61,6 +61,7 @@ from .custom_types import (
     _CUSTOM_Float,
     _CUSTOM_Time,
 )
+from .util import _update_connection_application_name
 
 colspecs = {
     Date: _CUSTOM_Date,
@@ -105,6 +106,9 @@ ischema_names = {
     "ARRAY": ARRAY,
     "GEOGRAPHY": GEOGRAPHY,
 }
+
+
+_ENABLE_SQLALCHEMY_AS_APPLICATION_NAME = True
 
 
 class SnowflakeDialect(default.DefaultDialect):
@@ -840,6 +844,18 @@ class SnowflakeDialect(default.DefaultDialect):
             if result and result._mapping["comment"]
             else None
         }
+
+    def connect(self, *cargs, **cparams):
+        return (
+            super().connect(
+                *cargs,
+                **_update_connection_application_name(**cparams)
+                if _ENABLE_SQLALCHEMY_AS_APPLICATION_NAME
+                else cparams,
+            )
+            if _ENABLE_SQLALCHEMY_AS_APPLICATION_NAME
+            else super().connect(*cargs, **cparams)
+        )
 
 
 @sa_vnt.listens_for(Table, "before_create")
--- a/src/snowflake/sqlalchemy/snowdialect.py
+++ b/src/snowflake/sqlalchemy/snowdialect.py
@@ -637,9 +637,6 @@ class SnowflakeDialect(default.DefaultDialect):
 
             type_instance = col_type(**col_type_kw)
 
-            if table_name not in schema_primary_keys:
-                raise sa_exc.NoSuchTableError()
-
             current_table_pks = schema_primary_keys.get(table_name)
 
             ans.append(
@@ -658,6 +655,10 @@ class SnowflakeDialect(default.DefaultDialect):
                     else False,
                 }
             )
+
+        # If we didn't find any columns for the table, the table doesn't exist.
+        if len(ans) == 0:
+            raise sa_exc.NoSuchTableError()
         return ans
 
     def get_columns(self, connection, table_name, schema=None, **kw):
--- a/src/snowflake/sqlalchemy/snowdialect.py
+++ b/src/snowflake/sqlalchemy/snowdialect.py
@@ -61,7 +61,6 @@ from .custom_types import (
     _CUSTOM_Float,
     _CUSTOM_Time,
 )
-from .util import _update_connection_application_name
 
 colspecs = {
     Date: _CUSTOM_Date,
@@ -834,12 +833,6 @@ class SnowflakeDialect(default.DefaultDialect):
             if result and result._mapping["comment"]
             else None
         }
-
-    def connect(self, *cargs, **cparams):
-        snowflake_conn = super().connect(
-            *cargs, **_update_connection_application_name(**cparams)
-        )
-        return snowflake_conn
 
 
 @sa_vnt.listens_for(Table, "before_create")
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -32,7 +32,7 @@ from sqlalchemy import (
     inspect,
     text,
 )
-from sqlalchemy.exc import DBAPIError
+from sqlalchemy.exc import DBAPIError, NoSuchTableError
 from sqlalchemy.pool import NullPool
 from sqlalchemy.sql import and_, not_, or_, select
 
@@ -415,6 +415,18 @@ def test_insert_tables(engine_testaccount):
             # drop tables
             addresses.drop(engine_testaccount)
             users.drop(engine_testaccount)
+
+
+def test_table_does_not_exist(engine_testaccount):
+    """
+    Tests Correct Exception Thrown When Table Does Not Exist
+    """
+    with engine_testaccount.connect() as conn:
+        meta = MetaData()
+        with pytest.raises(NoSuchTableError):
+            Table(
+                "does_not_exist", meta, autoload=True, autoload_with=engine_testaccount
+            )
 
 
 @pytest.mark.skip(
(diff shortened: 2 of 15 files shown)


davidov541:

davidov541:
recheck

sfc-gh-sfan:
@sfc-gh-aling : Do we consider this a behavior change? This falls into the discussion we had yesterday, where we correct a behavior.

davidov541:
Has there been any update on this? It's a pretty simple fix that enables a pretty powerful feature with Great Expectations (the ability to keep validation results in Snowflake directly for analysis), so the sooner this can get approved and merged, the sooner I can move my client off of using a custom build from my personal repository for this work.

sfc-gh-sfan (approved):
This looks good to me, I'll let @sfc-gh-aling chime in as well.

davidov541:
Great, thanks @sfc-gh-sfan . It looks like a few checks failed, but I don't see how they relate directly to my changes. Can you take a look and let me know if there are changes I need to make to make the checks work? The lint failure is below:

RuntimeError: The Poetry configuration is invalid:
            - [extras.pipfile_deprecated_finder.2] 'pip-shims<=0.3.4' does not match '^[a-zA-Z-_.0-9]+$'

src/snowflake/sqlalchemy/snowdialect.py
```diff
@@ -637,6 +637,9 @@ def _get_table_columns(self, connection, table_name, schema=None, **kw):
 
             type_instance = col_type(**col_type_kw)
 
+            if table_name not in schema_primary_keys:
+                raise sa_exc.NoSuchTableError()
```
sfc-gh-aling, line 641: may I ask why check is added here? the following line:
`current_table_pks = schema_primary_keys.get(table_name)` returns `None` if table not in `schema_primary_keys`.

also the added test doesn't hit the code path I think, do you have a scenario for it?

src/snowflake/sqlalchemy/snowdialect.py
```diff
@@ -669,7 +672,10 @@ def get_columns(self, connection, table_name, schema=None, **kw):
         if schema_columns is None:
             # Too many results, fall back to only query about single table
             return self._get_table_columns(connection, table_name, schema, **kw)
-        return schema_columns[self.normalize_name(table_name)]
+        normalized_table_name = self.normalize_name(table_name)
+        if normalized_table_name not in schema_columns:
+            raise sa_exc.NoSuchTableError()
+        return schema_columns[normalized_table_name]
```
sfc-gh-aling, line 678: could you help check and do us a favor to update if there're other similar usages of the `index` operator to get value from a dict which will trigger unexpected `KeyError` in `snowdialect.py`?

src/snowflake/sqlalchemy/snowdialect.py
```diff
@@ -669,7 +672,10 @@ def get_columns(self, connection, table_name, schema=None, **kw):
         if schema_columns is None:
             # Too many results, fall back to only query about single table
             return self._get_table_columns(connection, table_name, schema, **kw)
-        return schema_columns[self.normalize_name(table_name)]
+        normalized_table_name = self.normalize_name(table_name)
+        if normalized_table_name not in schema_columns:
+            raise sa_exc.NoSuchTableError()
```
sfc-gh-aling, line 677: we should document the changing behavior in the release notes that `NoSuchTableError` is raised instead of `KeyError`.

sfc-gh-aling (commented):
appreciate the PR! we would love to take it, may I ask you to do us a favor to see if there're any other places where `KeyError` should not be raised when checking the existence of tables (or other objects you find) and update the code accordingly. I've left comments, please let me know if you have any questions regarding my comments.

for the failure in CI, please try rebasing your branch to the latest main branch.

src/snowflake/sqlalchemy/snowdialect.py
davidov541: I made a new minor version in the description markdown file with today's date. If there is a different way you want to handle this let me know, but I see you edited it manually last week so I followed your pattern.

src/snowflake/sqlalchemy/snowdialect.py
davidov541: I did a review when I made this change for exactly this, but I just did another one, and here are all of the places I found that might lead to a key error. Let me know if any of these should be fixed and what error we should throw instead:

- L209: `host` is assumed to be in `opts` in `create_connect_args`
  - Not sure what we would want to throw here. 
- L329-336: A few keys are assumed to be returned by the query in `_get_schema_primary_keys`
  - This would be an internal error if it happened, so not sure a better error would be needed.
- L361-370: Same thing here in `_get_schema_unique_constraints`
- L401-423: Same thing again in `_get_schema_foreign_keys`
- L559: Dependency on internal API between `_get_schema_primary_keys` and `_get_schema_columns`. Shouldn't occur without a bug in the code.
- L655: Same here between `_get_schema_primary_keys` and `_get_table_columns`.
- L747: Another instance where we are depending on the results from the `connection.execute` call on `SHOW VIEWS...`. Regardless, exception is handled and ignored, so no KeyError would be shown to user.
- L769-670: Same thing here, but no catch of exceptions, so a KeyError would be shown to user. That said, it would be an internal bug if it happened, so not something we need to customize the error for.
- L838-839: Another instance of dependency on the result of a query, which would cause an internal error if it weren't true.

src/snowflake/sqlalchemy/snowdialect.py
davidov541: You're correct, but I think it would still cause an issue with L655, where we use the indexing operator on the same dictionary looking for the primary keys, which would throw a KeyError. So I think we still need it. I can add a comment on the check to clarify why we have it if that makes more sense?

davidov541: On your second question, we do not test it, but I think it would show up in the same use case as the other check, but only if there were too many columns for `_get_schema_columns` to return all of them. Then we fall into this function call on L674 and would hit this KeyError while returning, skipping the check on L676.

sfc-gh-aling: L655 `in schema_primary_keys[table_name]["constrained_columns"]` is only executed when L657 `if current_table_pks` is not None/False and `current_table_pks 
 = schema_primary_keys.get(table_name)` on L643. so L655 shouldn't be executed if `table_name` not in schema_primary_keys.

with the change, it actually changes the behavior as previously this `_get_table_columns` can return if table_name is None but now its raising Error.

src/snowflake/sqlalchemy/snowdialect.py
sfc-gh-aling: it looks awesome, thanks!

src/snowflake/sqlalchemy/snowdialect.py
davidov541: Ah OK, I see now. Thanks for explaining this.

I'm trying to replicate this code path to make sure the original bug and my original use case will still be fixed if we revert this piece. I've created 10,000 tables with 10,000 columns each in a schema, but it still doesn't seem to hit this code path, and I can't find any documentation online of the error 90030, which would lead to this code path. How would you suggest I reproduce this code path for testing purposes in my own account?

Thanks!

sfc-gh-aling:
appreciate the quick response and data collection!
apologize that I might not have the time to take a look this week due to some high priority tasks I have to deal with. I should have time next week to take a closer look at the code where we want to raise proper error.

davidov541:
> appreciate the quick response and data collection! apologize that I might not have the time to take a look this week due to some high priority tasks I have to deal with. I should have time next week to take a closer look at the code where we want to raise proper error.

No worries! I will be travelling for a week starting on Friday, so I'll be spotty myself, but any chance you get to give some feedback and get this resolved once and for all would be greatly appreciated!

davidov541:
> appreciate the quick response and data collection! apologize that I might not have the time to take a look this week due to some high priority tasks I have to deal with. I should have time next week to take a closer look at the code where we want to raise proper error.

Did you get any chance to look into this at all?

src/snowflake/sqlalchemy/snowdialect.py
sfc-gh-aling: thanks for collecting the usage. for the first one, I think this would be addressed by the PR: https://github.com/snowflakedb/snowflake-sqlalchemy/pull/387.

for the remaining ones, I concur with you that those should not require extra error handling.
again thanks for the efforsts

src/snowflake/sqlalchemy/snowdialect.py
sfc-gh-aling: I think I know why -- it's because this check is in for-looping the result, and when there is no table in the schema, the result is empty so the code path doesn't get hit as the for-loop doesn't get run at all.

To align the error experience, one thing we could do is to check whether `ans` is empty in the end -- if it's empty, then it means no table found in the schema. Do you want to try it on your end?

davidov541: @sfc-gh-aling - Thanks for this, I think that approach makes sense. I've gone ahead and made it, so please take a look and verify it is what you were expecting.

sfc-gh-aling: change looks good to me, thanks!

tests/test_core.py
```diff
@@ -406,6 +406,18 @@ def test_insert_tables(engine_testaccount):
...
+def test_table_does_not_exist(engine_testaccount):
+    """
+    Tests Correct Exception Thrown When Table Does Not Exist
+    """
+    with engine_testaccount.connect() as conn:
+        meta = MetaData()
+        with pytest.raises(NoSuchTableError):
+            Table(
+                "does_not_exist", meta, autoload=True, autoload_with=engine_testaccount
+            )
```
sfc-gh-aling, line 418:
```suggestion
    meta = MetaData()
    with pytest.raises(NoSuchTableError):
        Table("does_not_exist", meta, autoload=True, autoload_with=engine_testaccount)
```

sfc-gh-aling:
thanks @davidov541 , due to our ci limitation, I have created a draft PR for testing: https://github.com/snowflakedb/snowflake-sqlalchemy/pull/392/files

once it passes, let's get your PR merged. thank you so much for the contribution and apologize for my late response!

sfc-gh-aling (approved):

sfc-gh-aling:
hey @davidov541, do you mind if I merge the identical test PR instead of this PR?

We have some issue with the CI setting -- for now we don't allow credential to be included in a external contributor running, in the meanwhile we have branch protection rule that requires the checks to pass... it's an issue we need to fix, but to unblock you, I think we can merge the identical test PR.

davidov541:
> hey @davidov541, do you mind if I merge the identical test PR instead of this PR? (merge #392 and close this one)
> 
> We have some issue with the CI setting -- for now we don't allow credential to be included in a external PR test running so CIs are expected to fail, in the meanwhile we have branch protection rule that requires the checks to pass... it's an issue we need to fix, but to unblock you, I think we can merge the identical test PR.

I'm fine with you going with that approach if you all need it done.

PR #545, opened 2024-11-19
Added flag to allow override SnowflakeDialect div_is_floor_div flag

### LINKED ISSUE #543: SNOW-1812881: Division broken

1. What version of Python are you using? 3.12 (but irrelevant)
2. What operating system and processor architecture are you using? `macOS-14.7.1-arm64-arm-64bit` (but irrelevant)
3. What are the component versions in the environment (`pip freeze`)? The relevant ones are: `sqlalchemy 2.0.36` and `snowflake-sqlalchemy 1.6.1`
4. What did you do? Create a statement like `a / b` and it does not compile correctly.

```python
from snowflake.sqlalchemy.snowdialect import SnowflakeDialect
from sqlalchemy.sql.compiler import SQLCompiler
import sqlalchemy as sa

dialect = SnowflakeDialect()

stmt = sa.Column("a") / sa.func.sqrt(sa.Column("b"))
print(str(SQLCompiler(dialect, stmt)))
```

This prints `a / CAST(sqrt(b) AS NUMERIC)`. However, the cast in the denominator should not happen, and this cast leads to wrong results (!).

The fix is to set `div_is_floordiv = False` in `SnowflakeDialect`, see e.g. this as a test:

```python
from snowflake.sqlalchemy.snowdialect import SnowflakeDialect
from sqlalchemy.sql.compiler import SQLCompiler
import sqlalchemy as sa

class FixedDialect(SnowflakeDialect):
   div_is_floordiv = False

dialect = FixedDialect()

stmt = sa.Column("a") / sa.func.sqrt(sa.Column("b"))
print(str(SQLCompiler(dialect, stmt)))
```

BTW: fixing `div_is_floordiv` also fixes the integer division operator, i.e. `//`. Right now, this is also wrong:

```python
stmt = sa.Column("a", sa.Integer()) // sa.Column("b", sa.Integer())
print(str(SQLCompiler(dialect, stmt)))
```
currently prints "a / b". Using `FixedDialect` from above, it correctly prints `FLOOR(a / b)`.

sfc-gh-jmartinezramirez:

   Fixes #543 

   The default behavior of the flag div_is_floor_div was changed to False. This ensures that Python and Snowflake operators match the right division. In this first version a flag would be added `force_div_is_floor_div` to override the current behavior.

### DIFF
--- a/DESCRIPTION.md
+++ b/DESCRIPTION.md
@@ -10,6 +10,7 @@ Source code is also available at:
 # Release Notes
 - (Unreleased)
   - Fix return value of snowflake get_table_names
+  - Fix flag `div_is_floordiv` to `False` in `SnowflakeDialect` to avoid division wrong CAST when trying to do a integer division when using `/`
 
 - v1.7.2(December 18, 2024)
   - Fix quoting of `_` as column name
--- a/DESCRIPTION.md
+++ b/DESCRIPTION.md
@@ -10,7 +10,10 @@ Source code is also available at:
 # Release Notes
 - (Unreleased)
   - Fix return value of snowflake get_table_names
-  - Fix flag `div_is_floordiv` to `False` in `SnowflakeDialect` to avoid division wrong CAST when trying to do a integer division when using `/`
+  - Added `force_div_is_floordiv` flag to override `div_is_floordiv` new default value `False` in `SnowflakeDialect`.
+    - With the flag in `False`, the `/` division operator will be treated as a float division and `//` as a floor division.
+    - This flag is added to maintain backward compatibility with the previous behavior of Snowflake Dialect division.
+    - This flag will be removed in the future and Snowflake Dialect will use `div_is_floor_div` as `False`.
 
 - v1.7.2(December 18, 2024)
   - Fix quoting of `_` as column name
--- a/src/snowflake/sqlalchemy/snowdialect.py
+++ b/src/snowflake/sqlalchemy/snowdialect.py
@@ -148,7 +148,6 @@ class SnowflakeDialect(default.DefaultDialect):
 
     supports_identity_columns = True
 
-
     def __init__(
         self,
         force_div_is_floordiv: bool = True,
--- a/src/snowflake/sqlalchemy/snowdialect.py
+++ b/src/snowflake/sqlalchemy/snowdialect.py
@@ -148,12 +148,20 @@ class SnowflakeDialect(default.DefaultDialect):
 
     supports_identity_columns = True
 
+
     def __init__(
         self,
+        force_div_is_floordiv: bool = True,
         isolation_level: Optional[str] = SnowflakeIsolationLevel.READ_COMMITTED.value,
         **kwargs: Any,
     ):
         super().__init__(isolation_level=isolation_level, **kwargs)
+        self.force_div_is_floordiv = force_div_is_floordiv
+        self.div_is_floordiv = force_div_is_floordiv
+
+    def initialize(self, connection):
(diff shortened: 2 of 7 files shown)


sfc-gh-osalazarlizano (approved):

jochenott:
Looks good. You might want to enable these sqlalchemy-provided tests again, which I believe should pass now:
https://github.com/snowflakedb/snowflake-sqlalchemy/blob/main/tests/sqlalchemy_test_suite/test_suite_20.py#L84-L99

DESCRIPTION.md
```diff
@@ -9,7 +9,13 @@ Source code is also available at:
 
 # Release Notes
 
-- v1.7.0(November 22, 2024)
+- (Unreleased)
+  - Added `force_div_is_floordiv` flag to override `div_is_floordiv` new default value `False` in `SnowflakeDialect`.
+    - With the flag in `False`, the `/` division operator will be treated as a float division and `//` as a floor division.
```
jochenott, line 14: This is not how it works.

This flag is a signal from the dialect towards sqlalchemy that describes the behavior of this sql implementation (Snowflake). So if `select 4/5` returns 0, `div_is_floordiv` should be set to true. If it returns `0.8`, this flag must be set to false.

So the previous value was just a bug, which should be fixed. There is no need to make this configurable.

sfc-gh-jmartinezramirez: Hi @jochenott 

We agree that the correct value for the flag `div_is_floordiv` in Snowflake should be False. However, current users expect the `/ ` and `//` operators to work as true divisions, although `/ `should behave as a true division and `//`  as a floor division. That's why the behavioral change will be introduced as a flag, which will be removed in a future major release.

Thank you for pointing out this code issue and your contribution!!!

tests/test_core.py
```diff
@@ -1863,3 +1864,31 @@ def test_snowflake_sqlalchemy_as_valid_client_type():
         snowflake.connector.connection.DEFAULT_CONFIGURATION[
             "internal_application_version"
         ] = origin_internal_app_version
+
+
+@pytest.mark.feature_v20
+def test_division_force_div_is_floordiv_default():
+    engine = create_engine(URL(**CONNECTION_PARAMETERS))
```
sfc-gh-jvasquezrojas, line 1871: You can use the fixture `engine_testaccount` instead

tests/test_core.py
```diff
@@ -1863,3 +1864,31 @@ def test_snowflake_sqlalchemy_as_valid_client_type():
...
+                    select(literal(5) / literal(10), literal(5) // literal(10))
+                ).fetchall(),
+                [(0.5, 0.5)],
+            )
+
+
+@pytest.mark.feature_v20
+def test_division_force_div_is_floordiv_false():
+    engine = create_engine(
+        URL(**CONNECTION_PARAMETERS), **{"force_div_is_floordiv": False}
```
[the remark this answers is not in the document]
sfc-gh-jmartinezramirez: Done

src/snowflake/sqlalchemy/snowdialect.py
```diff
@@ -137,6 +140,18 @@ class SnowflakeDialect(default.DefaultDialect):
 
     supports_identity_columns = True
 
+    def __init__(
+        self,
+        force_div_is_floordiv_behavior: bool = True,
```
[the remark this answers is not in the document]
sfc-gh-jmartinezramirez: Done

jochenott:
Anything missing to get this merged and released?

PR #624, opened 2025-11-24
SNOW-2853075: Log SQLAlchemy version

sfc-gh-jcieslak:
## Changes
- Every time SQLAlchemy establishes a new connection, send the logs about the used SQLAlchemy version
- Add a unit test with mocks that validates the log is sent on new connections and doesn't break it when any of the logging action fails

### DIFF
--- a/DESCRIPTION.md
+++ b/DESCRIPTION.md
@@ -7,10 +7,12 @@ Snowflake Documentation is available at:
 Source code is also available at:
 <https://github.com/snowflakedb/snowflake-sqlalchemy>
 # Unreleased Notes
+- v1.7.8
+    - Add logging of SQLAlchemy version
 
 # Release Notes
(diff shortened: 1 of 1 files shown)


tests/test_dialect_connect.py
```diff
@@ -0,0 +1,98 @@
...
+from unittest.mock import MagicMock
+
+import pytest
+from sqlalchemy.engine import default as sqla_default
+
+from snowflake.sqlalchemy import snowdialect
+from snowflake.sqlalchemy.snowdialect import SnowflakeDialect, TelemetryField
+
+# @pytest.fixture(scope="session", autouse=True)
+# def init_test_schema():
```
sfc-gh-jcieslak: wasn't sure if that's needed, can remove/uncomment depending on how useful it is

sfc-gh-mraba (approved):

PR #449, opened 2023-09-25
SNOW-922868 Preserve the sequence order of foreign keys when dealing with composite keys

kaushalp-ai:
This MR has a fix for #450 (Preserve the sequence order of foreign keys when dealing with composite keys.)

I am reordering foreign keys based on the sequence key field.

PR #669, opened 2026-03-12
fix composite keys order

### LINKED ISSUE #450: Preserve the sequence order of foreign keys when dealing with composite keys.
Hi Team, 

I am encountering issues while working with composite foreign keys. For every version, I am receiving fixed create_foreign_key and drop_constraint operations. Upon debugging, I discovered that we are not preserving the sequence order of foreign keys when comparing with metadata.

For more detail, I am pointing out the specific code snippet that requires modification. - https://github.com/snowflakedb/snowflake-sqlalchemy/blob/main/src/snowflake/sqlalchemy/snowdialect.py#L414-L434C18

sfc-gh-jcieslak:

   Fixes #450
   References #449

   Adding sequence key to column queries so that in case of composite keys, they're being returned in order, which matters in cases like Alembic migrations (also following other dialects' behavior).


src/snowflake/sqlalchemy/snowdialect.py
```diff
@@ -480,20 +516,37 @@ def _get_schema_foreign_keys(self, connection, schema, **kw):
...
         ans = {}
-
         for _, v in foreign_key_map.items():
+            # Sort referred columns by key sequence
+            v["constrained_columns"] = [
+                col for _, col in sorted(v["constrained_columns"])
+            ]
+
+            # Sort referred columns by key sequence
+            v["referred_columns"] = [col for _, col in sorted(v["referred_columns"])]
```
sfc-gh-mraba: There are a few repetitions of this sorting.
Given the comment on lexicographic sort this could be replaced with helper 

```python
  @staticmethod
  def _sorted_by_key_sequence(pairs):
      """Sort (key_sequence, col_name) pairs and return col_names in order."""
      return [col for _, col in sorted(pairs, key=lambda x: int(x[0]))]

```

src/snowflake/sqlalchemy/snowdialect.py
```diff
@@ -407,24 +419,40 @@ def _get_schema_unique_constraints(self, connection, schema, **kw):
...
+                        self.normalize_name(row._mapping["column_name"]),
+                    )
                 )
 
         ans = defaultdict(list)
         for constraint in unique_constraints.values():
+            # Sort constrained columns by key sequence
+            constraint["column_names"] = [
+                col for _, col in sorted(constraint["column_names"])
+            ]
```
sfc-gh-mraba: `row._mapping["key_sequence"]` is likely returned as a string or Decimal by the Snowflake
  Python connector. Sorting tuples of (str, str) is lexicographic, which breaks for
  sequences beyond 9:
```
  "1", "10", "2"  ← string sort (wrong)
    1,   2,  10  ← int sort (correct)
```
  This is a latent bug that will only surface on tables with 10+ column composite keys —
  uncommon, but definitely possible in data-warehouse schemas.

src/snowflake/sqlalchemy/snowdialect.py
```diff
@@ -407,24 +419,40 @@ def _get_schema_unique_constraints(self, connection, schema, **kw):
...
+                    # Use pair (key_sequence, column_name) to sort constrained columns by key sequence
+                    (
+                        row._mapping["key_sequence"],
+                        self.normalize_name(row._mapping["column_name"]),
+                    )
                 )
 
         ans = defaultdict(list)
         for constraint in unique_constraints.values():
+            # Sort constrained columns by key sequence
```
sfc-gh-mraba: These are sorted by 'sequence' or 'column_names'?

sfc-gh-jcieslak: Adjusted within helper to make it more readable; it should be sorted by key_sequence

src/snowflake/sqlalchemy/snowdialect.py
sfc-gh-jcieslak: Good point, added helper

src/snowflake/sqlalchemy/snowdialect.py
sfc-gh-jcieslak: Good point, added NamedTuple which casts to int and helper works on this NamedTuple to ensure we are sorting on ints rather than strings

tests/test_keys.py
```diff
@@ -0,0 +1,172 @@
...
+        eq_(pk["constrained_columns"], ["col_a", "col_b", "col_c"])
+
+        fks = inspector.get_foreign_keys("test_keys_fk_child_decl")
+        eq_(len(fks), 1)
+        eq_(fks[0]["name"], "fk_test_keys_decl")
+        eq_(fks[0]["constrained_columns"], ["ref_a", "ref_b", "ref_c"])
+        eq_(fks[0]["referred_columns"], ["col_a", "col_b", "col_c"])
+        eq_(fks[0]["referred_table"], "test_keys_fk_parent_decl")
+    finally:
+        child.drop(engine_testaccount)
```
sfc-gh-asawicki: question: what if `child.drop` fails (e.g. throwing the exception)? parent.drop wouldn't be executed them, right? SHouldn't we use `metadata.drop_all(engine_testaccount)` instead in all such tests (example: https://github.com/snowflakedb/snowflake-sqlalchemy/blob/fc2e8faea964d93893fc01d713200ec4d40b0fa7/tests/test_core.py#L559)? (I see that it's also implemented like this in other tests, eg https://github.com/snowflakedb/snowflake-sqlalchemy/blob/fc2e8faea964d93893fc01d713200ec4d40b0fa7/tests/test_core.py#L326)

tests/test_keys.py
```diff
@@ -0,0 +1,172 @@
...
+    MetaData,
+    PrimaryKeyConstraint,
+    Table,
+    UniqueConstraint,
+    inspect,
+)
+from sqlalchemy.testing.assertions import eq_
+
+
+def test_composite_fk_reflects_key_order(engine_testaccount):
```
sfc-gh-asawicki: nit: these tests are pretty repeatable (they share similar setup, execution, and validation), I could see them as parametrized tests (it would make it easier to add more cases and focus on use cases). There are multiple options to implement paramterized tests, so maybe we could discuss later on.

sfc-gh-asawicki (approved):

tests/test_keys.py
sfc-gh-jcieslak: Done

src/snowflake/sqlalchemy/snowdialect.py
```diff
@@ -403,6 +412,12 @@ def get_check_constraints(self, connection, table_name, schema, **kw):
     # Shared row-parsing helpers
     # ---------------------------------------------------------------------------
 
+    @staticmethod
+    def _sort_columns_by_key_sequence(columns: list[_KeyedColumn]) -> list[str]:
+        """Sort columns by key_sequence and return column names."""
+        columns.sort(key=lambda c: c.key_sequence)
```
sfc-gh-mraba: `.sort()` mutates the caller's list. The method signature and docstring both look like a pure function that returns a transformed list, but it also has an invisible side effect. The callers are fine today because they immediately reassign the result but this could silently corrupt state if the list were ever passed from somewhere that also holds a reference. Use `sorted()` instead:

  ```python
return [c.column_name for c in sorted(columns, key=lambda c: c.key_sequence)]
```

  This is also more idiomatic Python for "return a sorted copy."

File: tests/test_keys.py
Comment (sfc-gh-mraba) [inline tests/test_keys.py:1]: `test_composite_fk_reflects_key_order` and  `test_composite_fk_when_parent_pk_order_differs_from_columns` both test multi-column FK reflection. The second one adds a twist (PK column order differs from table column order), which is the most important edge case. Consider merging the two, or making it clear in the first test's name that it only covers the "same order" case.

File: src/snowflake/sqlalchemy/snowdialect.py [hunk 7]
```diff
@@ -78,6 +78,15 @@ class SnowflakeIsolationLevel(Enum):
     AUTOCOMMIT = "AUTOCOMMIT"
 
 
+class _KeyedColumn(NamedTuple):
+    key_sequence: int
+    column_name: str
+
+    @classmethod
+    def new(cls, key_sequence, column_name: str) -> "_KeyedColumn":
+        return cls(int(key_sequence), column_name)
```
sfc-gh-mraba: I suggest to remove the factory entirely and use as 

```python
  _KeyedColumn(int(row._mapping["key_sequence"]), self.normalize_name(...))
```

  - The int() cast is now visible at the call site — it's explicit that the DB cursor returns a string and we're converting it. That's not something worth hiding.
  - Fewer moving parts. The factory saves no lines; every call site with `.new(x, y)` is the same length as `(int(x), y)`.
  - The `NamedTuple` constructor signature already documents the types `(key_sequence: int)`. A reader who sees `int(row._mapping["key_sequence"])` immediately understands the intent.
  - The `.new()` name is also misleading — it looks like it might do validation or have richer logic, but it's just a coercion wrapper.

  The only scenario where a factory would be justified is if key_sequence required non-trivial parsing (e.g., stripping units, handling nulls, mapping enums). A bare
  `int()` cast doesn't clear that bar.

src/snowflake/sqlalchemy/snowdialect.py
sfc-gh-jcieslak: Done

tests/test_keys.py
sfc-gh-jcieslak: Shortened by parametrizing

src/snowflake/sqlalchemy/snowdialect.py
sfc-gh-jcieslak: Done

sfc-gh-mraba (approved):

PR #670, opened 2026-04-15
Cross database joins

### LINKED ISSUE #456: Best practice for doing cross database joins using CORE
I am literally copying this from the ORM question because it's virtually identical but the ORM approach doesn't work for CORE.

Question

Hi there, in Snowflake we often have multiple databases, with their own set of schemas, and tables inside those schemas. Ideally data shouldn't have to move between these databases nor schemas, but in practice they often do.

One use case is when there might be valuable data between two different databases. In my example, I'm going to reference database `A` and database `B` which has schemas `ant` and `baboons`, which has tables `apples` and `bananas`, respectively.

Through a direct query using Snowflake SQL, you could implement a join like
```sql
select
    *
from A.ant.apples a
join B.baboons.bananas b
    on a.id = b.id
```
How can you implement a cross database join like the above example using a single session through the CORE interface? My understanding is that a session can only point towards one database. I'm not creating an app and the tables already exist so ideally this would all happen through SQLAlchemy MetaData object reflection. Below are some example of what I've tried.

The ideal approach would look like the following:

```python
import os

from sqlalchemy import create_engine, MetaData, text, Table
from snowflake.sqlalchemy import URL
from sqlalchemy.sql import select, and_, or_, case, func, insert


# Note: no database, schema and warehouse are being selected at this time. They're not needed yet.

engine = create_engine(URL(
    account="my_awesome_account",
    user="my_user",
    password="my_password",
    role="my_awesome_role",
    ),
    )
metadata = MetaData()

metadata.reflect(engine, schema='a.ant', only=['apples',]) # doesn't work
metadata.reflect(engine, schema='a.ant', only=['a.ant.apples',]) # doesn't work
metadata.reflect(engine, only=['a.ant.apples',]) # doesn't work

Table("apples", metadata, schema='a.ant', autoload_with=engine) # doesn't work
Table("a.ant.apples", metadata, schema='a.ant', autoload_with=engine) # doesn't work
```

I've tried all types of combinations of cases with the the database name and schema names too. None of them work. 
What I find silly about the above is I often get StackTraces saying I have to provide a fully qualified name since I'm not declaring a database. But I _AM_ providing a fully qualified name. And not explicitly choosing a database is the point. I need to pull data from tables in multiple databases so the reflected tables _NEED TO BE FULLY QUALIFIED_.

Can this be done through the CORE interface without dropping to SQL? If so, what's the best practice in doing this?

   `Python 3.9.6`

   `macOS-14.0-arm64-arm-64bit`

   `appnope==0.1.3
asn1crypto==1.5.1
asttokens==2.4.0
backcall==0.2.0
certifi==2023.7.22
cffi==1.16.0
charset-normalizer==3.3.0
comm==0.1.4
cryptography==41.0.4
debugpy==1.8.0
decorator==5.1.1
exceptiongroup==1.1.3
executing==2.0.0
filelock==3.12.4
idna==3.4
importlib-metadata==6.8.0
ipykernel==6.25.2
ipython==8.16.1
jaraco.classes==3.3.0
jedi==0.19.1
jupyter_client==8.4.0
jupyter_core==5.4.0
keyring==24.2.0
matplotlib-inline==0.1.6
more-itertools==10.1.0
nest-asyncio==1.5.8
numpy==1.26.1
oscrypto==1.3.0
packaging==23.2
pandas==2.0.3
parso==0.8.3
pexpect==4.8.0
pickleshare==0.7.5
platformdirs==3.11.0
prompt-toolkit==3.0.39
psutil==5.9.6
ptyprocess==0.7.0
pure-eval==0.2.2
pyarrow==10.0.1
pycparser==2.21
pycryptodomex==3.19.0
Pygments==2.16.1
PyJWT==2.8.0
pyOpenSSL==23.2.0
python-dateutil==2.8.2
pytz==2023.3.post1
pyzmq==25.1.1
requests==2.31.0
six==1.16.0
snowflake-connector-python==3.3.1
snowflake-sqlalchemy==1.5.0
sortedcontainers==2.4.0
SQLAlchemy==1.4.49
sqlparse==0.4.4
stack-data==0.6.3
tomlkit==0.12.1
tornado==6.3.3
traitlets==5.11.2
typing_extensions==4.8.0
tzdata==2023.3
urllib3==1.26.18
wcwidth==0.2.8
zipp==3.17.0
`

4. What did you do?

   The above example should be sufficient.

   Ideally, the fully qualified table objects, including the database, would have been instantiated as SQLAlchemy Table objects. In this way, I could then create a SQLAlchemy `select` statement such as `select(a).select_from(a.join(b, on=[a.c.col1 == b.c.col1])` and the corresponding SQL generated would be `SELECT A.ant.apple.* FROM A.ant.apple JOIN B.baboons.bananas ON A.ant.apple.col1 = B.baboons.bananas.col1;`.

Instead, the reflection step fails forcing me to choose a database which then prevents me from joining tables in different databases.

6. Can you set logging to DEBUG and collect the logs?

   ```
   import logging
   import os

   for logger_name in ['snowflake.sqlalchemy', 'snowflake.connector']:
      logger = logging.getLogger(logger_name)
      logger.setLevel(logging.DEBUG)
      ch = logging.StreamHandler()
      ch.setLevel(logging.DEBUG)
      ch.setFormatter(logging.Formatter('%(asctime)s - %(threadName)s %(filename)s:%(lineno)d - %(funcName)s() - %(levelname)s - %(message)s'))
      logger.addHandler(ch)
   ```

I'm happy to provide this info but in private. It's on a company computer with company backend stuff that's not for the public eye.

sfc-gh-jcieslak:

   Fixes #456

   Add the possibility to create cross-database joins by adjusting how the schema is parsed.


sfc-gh-mraba (commented):
Pls consider pausing this change and applying it after #675 is reviewed and accepted.
After rebasing following steps should bring this solution to life - I'd be happy to jump in:
- `_split_schema_by_dot` can be dropped - already present in #675 
-  Restore same-schema FK detection (the critical regression)
The root cause: `_get_schema_foreign_keys` now receives `full_schema_name`  (`"DB"."SCHEMA"`) as its schema parameter. `normalize_name` on a composite string never equals a plain `pk_schema`, so the user's schema is never in `same_schemas`.

  The right fix is to separate SQL identity (what goes into SHOW) from the FK comparison identity (what to compare normalized pk_schema against). The cleanest way without changing the
  cached function's cache key:
```python
  # In get_foreign_keys / get_multi_foreign_keys — pass raw schema through
  def get_foreign_keys(self, connection, table_name, schema=None, **kw):
      schema = schema or self.default_schema_name
      if self._is_single_table_reflection(schema, **kw):
          return self._get_table_foreign_keys(connection, table_name, schema, **kw)
      full_schema_name = self._get_full_schema_name(connection, schema, **kw)
      # Pass both: full_schema_name for the SHOW SQL,
      # schema (raw) for same-schema FK detection
      return self._get_schema_foreign_keys(
          connection, full_schema_name, user_schema=schema, **kw
      ).get(table_name, [])

  @reflection.cache
  def _get_schema_foreign_keys(self, connection, schema, user_schema=None, **kw):
      current_database, current_schema = self._current_database_schema(connection, **kw)
      same_schemas = {self.default_schema_name, current_schema}

      # Add the user-specified schema (schema part only, not db prefix)
      # so same-schema FKs within a non-default schema get referred_schema=None
      if user_schema:
          parts = self.identifier_preparer._split_schema_by_dot(user_schema)
          schema_only = parts[-1]          # last part is always the schema
          if "." not in str(schema_only):  # guard: don't add composite strings
              same_schemas.add(self.normalize_name(str(schema_only)))

      result = connection.execute(
          text(f"SHOW /* sqlalchemy:_get_schema_foreign_keys */ IMPORTED KEYS IN SCHEMA {schema}")
      )
      return self._parse_fk_rows(result, same_schemas, current_database)
```
  Apply the same fix to `_get_table_foreign_keys` (which already receives the raw `schema`; it just needs to stop discarding it from `same_schemas` when the schema is cross-database — the
  current PR #670 logic of `if "." not in normalized_schema` is correct there, just the composite-string-as-input problem is unique to the schema-wide path).

  ▎ Why this matters: the existing docstring in _get_schema_foreign_keys explicitly states that the reflected schema must be in same_schemas for non-default-schema reflection to work. PR
  #670 breaks this invariant silently.
- Quote referred_schema in `_parse_fk_rows` (small, defensive)
- Remove double-parse in `_query_all_columns_info` (optional but recommended)
- Add one regression test
```python
  def test_get_foreign_keys_non_current_schema_same_db(engine_testaccount, ...):
      """FKs within a non-current schema should have referred_schema=None."""
      # Create a schema that differs from current_schema
      # Create two tables with an FK between them in that schema
      # Reflect FKs, assert referred_schema is None
```
  This would have caught the same_schemas regression.

sfc-gh-mraba (approved):

PR #671, opened 2026-04-15
Fix alembic multi schema migrations

### LINKED ISSUE #610: SNOW-2313675: Alembic trying to create same migrations when using multi schema migrations

   python3.12

   macOS-15.5-x86_64-i386-64bit

   snowflake-connector-python==3.17.2
   snowflake-sqlalchemy==1.7.6
   SQLAlchemy==2.0.43

4. What did you do?

   Trying to add alembic for version support. We have two schemas per database and using one schema to track the migrations and few tables having foreign key mappings between these schemas. Upon creating first migration, it created tables fine. When I add new columns to the tables, it is again creating tables in the new migration file but not detecting the changes columns only.

   When creating a new migration, it should show only the changed columns

sfc-gh-jcieslak:

   Fixes #610

    Address a bug in foreign key reflection related to Alembic migrations


DESCRIPTION.md
```diff
@@ -9,6 +9,7 @@ Source code is also available at:
 
 # Unreleased Notes
 
+- Fix foreign key reflection returning wrong `referred_schema` for same-schema FKs in non-default schemas, which caused Alembic to generate spurious migrations ([#610](https://github.com/snowflakedb/snowflake-sqlalchemy/issues/610), SNOW-2313675).
```
sfc-gh-mraba: No source fix in this PR; fix is already in main from #656 — PR description is misleading

tests/test_core.py
sfc-gh-mraba, line 1: `DROP SCHEMA` without `CASCADE` in two tests

  `test_get_foreign_keys_multi_schema`:
```python
  # finally block:
  metadata.drop_all(engine_testaccount)          # drops tables
  conn.execute(text(f"DROP SCHEMA IF EXISTS {schema1_name}"))   # no CASCADE!
  conn.execute(text(f"DROP SCHEMA IF EXISTS {schema2_name}"))   # no CASCADE!
```

  If drop_all fails partway, the subsequent DROP SCHEMA will fail (schema not empty), leaving both schemas behind.

  `test_alembic_autogenerate_multi_schema_fk` has the same problem — tables are created in schema1/schema2 but schemas are dropped without `CASCADE`.

```python
  test_reflection_same_schema_fk does use CASCADE:
  conn.execute(text(f"DROP SCHEMA IF EXISTS {schema_name} CASCADE"))
```

Use `CASCADE` or ensure tables are explicitly dropped first in a nested `try/finally`.

tests/alembic_integration/test_multi_schema_fk.py
sfc-gh-mraba, line 1: `get_multi_foreign_keys` (SA 2.x bulk path) not tested

  `test_get_foreign_keys_multi_schema` calls `inspector.get_foreign_keys(...)` which in SA 2.x typically goes through the single-table path (`get_foreign_keys`). The `get_multi_foreign_keys` bulk path (used by `MetaData.reflect()`) is a different code path and not directly exercised by these tests. The alembic integration test does test `MetaData.reflect()` behaviour indirectly, but a targeted unit test for `get_multi_foreign_keys` would be more precise.

tests/test_core.py
```diff
@@ -740,6 +741,150 @@ def test_get_foreign_keys(engine_testaccount):
...
+            foreign_keys_default = inspector.get_foreign_keys(
+                "items", schema=schema2_name
+            )
+            assert len(foreign_keys_default) == 1
+            assert foreign_keys_default[0]["name"] == "fk_to_default"
+            assert foreign_keys_default[0]["referred_table"] == "categories"
+            assert foreign_keys_default[0]["referred_schema"] is None
+
+        finally:
+            metadata.drop_all(engine_testaccount)
```
sfc-gh-mraba: Two separate connections are used concurrently. This is harmless since Snowflake DDL is transaction-independent, but the pattern is unusual and could confuse readers.
   The typical pattern in this test suite is to manage everything through one connection. Not a blocker, but worth tidying.

sfc-gh-mraba (commented):
`referred_schema=None` for cross-schema FK to default schema may cause Alembic mismatch

  `snowdialect.py` lines 716–720 (`_get_table_foreign_keys`) and 753–757 (`_get_schema_foreign_keys`):

  ```python
  same_schemas = {
      self.normalize_name(schema),  # the schema being reflected
      self.default_schema_name,     # ← this line
      current_schema,
  }

  default_schema_name is included so that FKs pointing at the default schema return referred_schema=None (no qualifier needed). However, when reflecting a non-default
  schema (schema2) that has a FK to the default schema (public), this produces referred_schema=None even though the source and target schemas differ.

  Consequence for Alembic autogenerate: if the application's MetaData defines the FK with an explicit schema —

  ForeignKey(f"{default_schema}.categories.id", name="fk_to_default")

  — Alembic will see referred_schema="public" in metadata vs. referred_schema=None from reflection and generate a spurious remove_fk / add_fk migration.

  The new test asserts the current None behavior:

  assert foreign_keys_default[0]["referred_schema"] is None

  but there is no Alembic autogenerate round-trip test that covers this exact case (non-default-schema table → default-schema table), so the potential mismatch is
  untested. Consider either:

  1. Adding an Alembic autogenerate test for this scenario (FK from schema2 to the default schema, where metadata defines the FK with an explicit schema), or
  2. Documenting intentionally that referred_schema=None is the expected value for FK targets in the default schema, so consumers know to define their ForeignKey(...)
  without a schema qualifier in that case.

sfc-gh-mraba (approved):

PR #675, opened 2026-04-17
SNOW-1232488 feat case sensitive support

sfc-gh-mraba:

   Fixes SNOW-1232488

  Fix case-sensitive identifier handling (SNOW-1232488)

  Fixes #[issue]

  Background

  Snowflake stores unquoted identifiers in uppercase (`mytable → MYTABLE`). SQLAlchemy's dialect normalises names on the way in (`MYTABLE → mytable`) and denormalises on the way out (`mytable → MYTABLE`). Three bugs in this round-trip silently corrupted case-sensitive identifiers, and there was no supported way to opt into correct handling of mixed-case quoted names.

  What changed

  Bug fixes (always active):

  - `denormalize_column_name` in `base.py` — now correctly emits double-quoted SQL for `quoted_name("mycol", True)` columns in `CLUSTER BY` clauses. Previously the `quote=True` signal was dropped and the column was silently rendered unquoted.
  - `_has_object in snowdialect.py` — now applies denormalize_name before building the `DESC TABLE` statement, making `has_table` / `has_sequence` consistent with all other reflection methods.
  - `_split_schema_by_dot` in `base.py` — now preserves `quote=True` on inner-quoted schema parts during dot-splitting, preventing normalisation loss for schemas like `"myDb"."mySchema"`.

 ### New opt-in flag — `case_sensitive_identifiers`:

  Pass as a `create_engine` keyword or `?case_sensitive_identifiers=true` URL parameter. When enabled, ALL-UPPERCASE reserved-word identifiers (e.g. a column literally named `TABLE`) are normalised to `quoted_name("table", True)` instead of being returned unchanged, preventing key-lookup mismatches between object creation and reflection.

  New helpers:

  - `create_snowflake_engine(url, schema=..., case_sensitive_schema=True)` — URL-encodes case-sensitive schema names via `%22` so the Snowflake connector receives the correct double-quoted form.
  - `snowflake.sqlalchemy.alembic_util.render_item` — a drop-in Alembic `render_item` hook for `env.py` that serialises `quoted_name` columns correctly in generated migration files, preventing autogenerate from silently uppercasing case-sensitive column names.

  Test infrastructure:

  - 54 unit tests covering all new and fixed code paths (mock-based, no live connection required).
  - `--case-sensitive` pytest flag to run the entire integration suite with the flag enabled for local debugging.
  - `engine_testaccount_case_sensitive` fixture for targeted case-sensitive integration tests in CI.
  - Fixed flaky `test_begin_read_commited` — buffer flush after `create_all` prevents DDL COMMITs from polluting transaction count assertions.

  Scope

  No changes to default normalisation behaviour. All fixes behind the `case_sensitive_identifiers` flag are additive. Existing applications are unaffected unless they explicitly opt in.


README.md
```diff
@@ -1094,3 +1095,318 @@ with Session(engine) as session:
...
+    else:
+        print(f"{name!r} — case-insensitive (Snowflake stores as {name.upper()})")
+```
+
+This means that after reflection, accessing a case-insensitive column requires the lowercase name:
+
+```python
+metadata.reflect(bind=engine)
+t = metadata.tables["my_table"]
+t.c.mycol      # correct — Snowflake stored MYCOL, reflected as "mycol"
```
sfc-gh-jcieslak, line 1139: This lacks information about what table and column we're talking about here. Please, specify. As I understand we're talking here about this case `create table my_table (mycol <type>)`?

I guess this exact example could be specified below the table with different quoting versions to set the context for all below sections.

README.md
```diff
@@ -1094,3 +1095,318 @@ with Session(engine) as session:
...
+The most reliable way to declare a case-sensitive schema across multiple tables is `MetaData(schema=quoted_name(..., True))`.  All tables and ORM models that share that `MetaData` object inherit the schema automatically — no repetition per table:
+
+```python
+from sqlalchemy import Column, Integer, MetaData, String, Table
+from sqlalchemy.sql.elements import quoted_name
+
+# Declare once — every Table attached to this MetaData inherits it.
+metadata = MetaData(schema=quoted_name("myschema", True))
+
+orders = Table("orders", metadata, Column("id", Integer, primary_key=True))
```
sfc-gh-jcieslak, line 1209: Question: Shouldn't orders in this case be wrapped in quoted_name to result in case sensitive table name?

README.md
```diff
@@ -1094,3 +1095,318 @@ with Session(engine) as session:
...
+```
+
+Plain strings with inner double-quotes — `'"myschema"'` — are also accepted and produce the same result as `quoted_name("myschema", True)`:
+
+```python
+Table("orders", metadata, schema='"myschema"')         # also works
+Table("orders", metadata, schema='"mydb"."myschema"')  # dotted form also works
+```
+
+The `"` characters are treated as SQL quoting delimiters by the schema parser, so the extracted name receives `quote=True` automatically.  Prefer `quoted_name` in code that is read often; prefer the quoted-string form when the value comes from a config that already embeds the quotes.
```
sfc-gh-jcieslak, line 1280: Quotes can cancel out when they're next to each other so `some""name` should be one part. Let's make sure it's handled properly or documented as exception

README.md
```diff
@@ -1094,3 +1095,318 @@ with Session(engine) as session:
...
+    __table_args__ = {"schema": "myschema"}  # also wrong
+```
+
+A plain lowercase string has no `quote` attribute, so `denormalize_name` converts it to `MYSCHEMA` before it reaches Snowflake.  No error is raised — if a schema named `MYSCHEMA` happens to exist the query succeeds silently against the wrong schema.
+
+Use `quoted_name("myschema", True)`, `'"myschema"'`, or `MetaData(schema=quoted_name(..., True))` as shown in the section above.
+
+---
+
+**Alembic autogenerate without `render_item` for case-sensitive columns**
```
sfc-gh-jcieslak: Is this necessary? It was already mentioned above that without the hook it won't be rendered well

tests/test_transactions.py
```diff
@@ -58,6 +58,10 @@ def test_begin_read_commited(engine_testaccount, assert_text_in_buf):
 
     metadata.create_all(engine_testaccount)
     try:
+        # Flush buffer so setup COMMITs (from create_all / _has_object)
```
sfc-gh-jcieslak: I don't think we need comments like this. Adjusting the name and self-commenting in GH would be sufficient. Now, looking in come time we won't know why such comment exist in such place. Plz remove.

src/snowflake/sqlalchemy/alembic_util.py
```diff
@@ -0,0 +1,93 @@
...
+        parts = [f"sa.sql.elements.quoted_name({col_name!r}, True)", rendered_type]
+
+        # Add important column attributes that differ from defaults
+        # nullable defaults to True, so only emit nullable=False
+        if not obj.nullable:
+            parts.append("nullable=False")
+
+        # Emit primary_key when True
+        if obj.primary_key:
+            parts.append("primary_key=True")
```
sfc-gh-jcieslak, line 121: This feels very fragile, isn't there a way to influence only identifier name and leave alembic the rest of the job to map those column attributes?

src/snowflake/sqlalchemy/name_utils.py
sfc-gh-jcieslak: This means the identifier contains mixed-casing, shouldn't we in this case also return quoted_name with quote = True?

tests/test_unit_case_sensitivity.py
```diff
@@ -0,0 +1,939 @@
...
+            ip.quote_schema(quoted_name("mydb.myschema", True)) == '"mydb"."myschema"'
+        )
+
+    # ------------------------------------------------------------------
+    # Plain string with inner double-quotes — these FAIL before the fix
+    # ------------------------------------------------------------------
+
+    def test_inner_quoted_single_schema(self, ip):
+        """'"myschema"' must produce '"myschema"' in SQL, not 'myschema'."""
+        assert ip.quote_schema('"myschema"') == '"myschema"'
```
sfc-gh-jcieslak, line 77: maybe it's somewhere else, but we're missing `'"my""schema"'` case

tests/test_unit_case_sensitivity.py
```diff
@@ -0,0 +1,939 @@
...
+    def test_split_parts_dotted_mixed(self, ip):
+        """'"mydb".myschema' → first part quote=True, second quote=None."""
+        parts = ip._split_schema_by_dot('"mydb".myschema')
+        assert len(parts) == 2
+        assert getattr(parts[0], "quote", None) is True
+        assert getattr(parts[1], "quote", None) is None
+
+
+# ---------------------------------------------------------------------------
+# Bug 1 — denormalize_column_name respects quoted_name.quote
```
sfc-gh-jcieslak: nit: We could remove "Bug 1", it's meaningless

tests/test_unit_case_sensitivity.py
```diff
@@ -0,0 +1,939 @@
...
+        compiler_cls = dialect.ddl_compiler
+        # Create a minimal object: DDLCompiler requires (dialect, statement, ...)
+        # Using a real table's create statement as the "statement"
+        from sqlalchemy.schema import CreateTable
+
+        stmt = CreateTable(t)
+        compiler = compiler_cls(dialect, stmt)
+        return compiler
+
+    def test_quoted_name_true_returns_quoted(self):
```
sfc-gh-jcieslak: Let's turn those tests into parameterized test, it would be easier to read and maintain

tests/test_unit_case_sensitivity.py
```diff
@@ -0,0 +1,939 @@
...
+
+    def test_quoted_name_false_falls_through(self):
+        """quoted_name('mycol', False) should follow normal branch (no forced quote)."""
+        compiler = self._get_compiler()
+        # quote=False means "let the dialect decide" — for simple lowercase it
+        # should come back unquoted.
+        result = compiler.denormalize_column_name(quoted_name("mycol", False))
+        assert result == "mycol", f"Expected 'mycol', got {result!r}"
+
+    def test_round_trip_normalize_denormalize_with_flag_on(self):
```
sfc-gh-jcieslak: imo we should have more round trip normalize denormalize test cases. Let's add all cases as parametrized test

tests/test_unit_case_sensitivity.py
```diff
@@ -0,0 +1,939 @@
...
+# ---------------------------------------------------------------------------
+
+
+class TestNormalizeNameReservedWord:
+    """Bug 2: normalize_name('TABLE') with flag on vs off."""
+
+    def _dialect(self, case_sensitive=False):
+        return SnowflakeDialect(case_sensitive_identifiers=case_sensitive)
+
+    def test_reserved_word_flag_off_returns_uppercase(self):
```
sfc-gh-jcieslak: Those tests also seem like could be parametrized (same as other tests below as I can see)

tests/test_unit_case_sensitivity.py
```diff
@@ -0,0 +1,939 @@
...
+        ), f"Expected 'mycol' in inner dict, got keys: {list(inner.keys())}"
+
+
+# ---------------------------------------------------------------------------
+# CLUSTER BY DDL — quoted_name column generates correct SQL
+# ---------------------------------------------------------------------------
+
+
+class TestClusterByWithQuotedName:
+    """Bug 1 (compiler path): CLUSTER BY DDL with quoted_name column."""
```
sfc-gh-jcieslak: If that's referring to a particular GitHub issue, let's link it here

src/snowflake/sqlalchemy/name_utils.py
sfc-gh-mraba: Valid. The else branch in `normalize_name` returns mixed-case names (e.g. `"MyTable"`) as plain strings. But in Snowflake, a mixed-case name can only exist if it was SQL-quoted (`CREATE TABLE "MyTable"`), so it is by definition case-sensitive. Returning `quoted_name("MyTable", True)` is correct and fixes a real semantic gap.

tests/test_unit_case_sensitivity.py
sfc-gh-mraba: Valid. The double-quote escape in SQL (`"my""schema"` = identifier named `my"schema`) is not tested. Running the parser mentally against `"my""schema"`: it would split into two parts `["my", "schema"]` instead of one. This is a parser bug. 
Added new tests 
- `test_inner_quoted_embeded_quote_roundtrip`
- `test_split_parts_embedded_quote_single_part`

src/snowflake/sqlalchemy/alembic_util.py
sfc-gh-mraba: Alembic's `_render_column` formats the name slot as `%(name)r`, which calls `repr()` — so by injecting `_ReprExpr` as `column.name`, we get the right expression in the output while Alembic handles every other attribute natively. Anti-recursion is handled by blanking `opts["render_item"]` before the call and restoring in `finally`. The manual reconstruction is kept as a fallback for environments where Alembic's `internal _render_column` is unavailable.

tests/test_unit_case_sensitivity.py
sfc-gh-mraba: Kept the round-trip test separate

README.md
sfc-gh-mraba: 1. We have table `my_table` with single column created with SQL
2. reflection on engine
3. `my_table` accessed explicitly

README.md
sfc-gh-mraba: Clarifying the semantics: `""` inside a quoted identifier is the SQL standard escape for a literal `"` character (reduction to one `"`, not cancellation to nothing). So `"my""schema"` is a single identifier named `my"schema`. This is documented in Snowflake's identifier syntax reference:
  https://docs.snowflake.com/en/sql-reference/identifiers-syntax#double-quoted-identifiers.

Added test and docs.
- `TestSplitSchemaByDot.test_inner_quoted_embedded_quote_roundtrip`
- `TestSplitSchemaByDot.test_split_parts_embedded_quote_single_part`

src/snowflake/sqlalchemy/alembic_util.py
```diff
@@ -54,40 +77,64 @@ def render_item(type_, obj, autogen_context):
...
+
+                rendered_default = _render_server_default(
+                    obj.server_default, autogen_context
+                )
                 parts.append(f"server_default={rendered_default}")
-            except (KeyError, AttributeError):
-                # Fall back to repr if renderer not available
-                parts.append(f"server_default={repr(obj.server_default)}")
+            except Exception:
+                # repr(DefaultClause) is not valid Python; emit a placeholder
```
sfc-gh-jcieslak: nit: this comment comes from the code that was previously here, idk if it makes much sense with the current code

tests/test_unit_case_sensitivity.py
```diff
@@ -119,160 +127,185 @@ def test_split_parts_dotted_mixed(self, ip):
...
+            # Mixed-case: can only exist as a SQL-quoted identifier → case-sensitive
+            ("MyTable", False, "MyTable", True),
+            ("MyTable", True, "MyTable", True),
+            # ALL-UPPERCASE reserved word: flag=False leaves unchanged (legacy)
+            ("TABLE", False, "TABLE", None),
+            # ALL-UPPERCASE reserved word: flag=True converts to quoted_name
+            ("TABLE", True, "table", True),
+            # Edge cases
+            (None, False, None, None),
+            ("", False, "", None),
```
sfc-gh-jcieslak, line 269: nit: missing edge cases for flag = True

tests/test_unit_case_sensitivity.py
```diff
@@ -119,160 +127,185 @@ def test_split_parts_dotted_mixed(self, ip):
...
+            ("MYTABLE", False, "MYTABLE"),
+            ("MYTABLE", True, "MYTABLE"),
+            # Quoted lowercase: normalizes to quoted_name, denormalizes to double-quoted SQL
+            ("mytable", False, '"mytable"'),
+            ("mytable", True, '"mytable"'),
+            # Mixed-case: normalizes to quoted_name, denormalizes to double-quoted SQL
+            ("MyTable", False, '"MyTable"'),
+            ("MyTable", True, '"MyTable"'),
+            # Reserved word with flag on: normalizes to quoted_name, denormalizes to double-quoted
+            ("TABLE", True, '"table"'),
```
sfc-gh-jcieslak: nit: missing case for flag = False

tests/test_unit_case_sensitivity.py
```diff
@@ -119,160 +127,185 @@ def test_split_parts_dotted_mixed(self, ip):
...
+        ],
+    )
+    def test_normalize_denormalize_round_trip(self, snowflake_name, flag, expected_sql):
+        """normalize_name → denormalize_name round-trip produces the expected SQL identifier."""
+        d = self._dialect(case_sensitive=flag)
+        normalized = d.normalize_name(snowflake_name)
+        result = d.denormalize_name(normalized)
+        # The identifier preparer would then quote result if needed; check the
+        # denormalized string directly to verify the round-trip value.
+        assert str(result) == expected_sql.strip('"') or result == expected_sql.strip(
```
sfc-gh-jcieslak: Question: Should we strip quotes in this assertion? And do we need two conditions in this case? I'm not sure what cases `result` vs `str(result)` would cover

tests/test_unit_case_sensitivity.py
```diff
@@ -751,14 +838,20 @@ def test_quoted_name_column_with_server_default(self):
 
         from snowflake.sqlalchemy.alembic_util import render_item
 
+        try:
+            from alembic.autogenerate.render import _render_column  # noqa: F401
+        except ImportError:
+            pytest.skip("alembic not installed")
```
sfc-gh-jcieslak: I think alembic may be missing from pyproject where it should be added as dev / test depedency. Imo we should remove those skips and have alembic as required package (as mentioned, for dev / test profiles).

src/snowflake/sqlalchemy/name_utils.py
```diff
@@ -36,7 +36,10 @@ def normalize_name(self, name):
         elif name.lower() == name:
             return quoted_name(name, quote=True)
         else:
-            return name
+            # Mixed-case names (e.g. "MyTable") can only exist in Snowflake when
+            # the identifier was SQL-quoted at creation time, so they are always
+            # case-sensitive.
+            return quoted_name(name, quote=True)
```
sfc-gh-jcieslak, line 45: Question: shouldn't we check for case_sensitive flag here and return quoted_name version only if flag is set and otherwise return `name` for backward compatibility?

sfc-gh-mraba: The flag controls all-uppercase reserved-word identifiers, where both the old behaviour (return unchanged) and the new behaviour (return quoted_name) are plausible interpretations. Mixed-case is categorically different: Snowflake cannot store a mixed-case identifier without SQL quoting it at creation time. If reflection returns `"MyTable"`, it came from a `"MyTable"` column and is inherently case-sensitive. Gating this on the flag would silently break round-trips for anyone who has mixed-case columns but hasn't opted in, which is the worse failure mode.

tests/test_unit_case_sensitivity.py
sfc-gh-mraba: Added alembic to `[project.optional-dependencies] development` in `pyproject.toml` and removed all `pytest.skip("alembic not installed")` guards.

tests/test_unit_case_sensitivity.py
sfc-gh-mraba: Fixed. 
`str(result) == X` or `result == X` is redundant since `quoted_name.__eq__` compares by value. Simplified to `assert result == expected_raw` and renamed the parameter column `expected_sql` → `expected_raw` to clarify that `denormalize_name` returns the raw string value, not a SQL-quoted form.

tests/test_unit_case_sensitivity.py
sfc-gh-mraba: Added `("TABLE", False, "TABLE")` — reserved word with `flag=False` normalizes unchanged and denormalizes back to `"TABLE"`.

tests/test_unit_case_sensitivity.py
sfc-gh-mraba: Fixed. Added `(None, True, None, None)` and `("", True, "", None)`.

src/snowflake/sqlalchemy/alembic_util.py
sfc-gh-mraba: Reworded to describe the current code directly: `sa.text(...)` is emitted as the minimum valid Python expression for a server default when `_render_server_default` is unavailable.

sfc-gh-jcieslak (approved):

sfc-gh-jcieslak (approved):

PR #677, opened 2026-04-20
Fix global CompileState plugin_for breaking with_loader_criteria on non-Snowflake dialects

### LINKED ISSUE #676: CompileState.plugin_for global registration breaks with_loader_criteria on non-Snowflake dialects
## Environment

1. **Python version:** 3.11+ (reproduced on 3.11, 3.12, 3.14)

2. **OS/architecture:** Reproducible on all platforms — this is a compile-time (SQL generation) bug, not platform-specific.

3. **Component versions:**
   ```
   snowflake-sqlalchemy==1.9.0
   sqlalchemy==2.0.49
   ```

## 4. What did you do?

Importing `snowflake.sqlalchemy` anywhere in the process silently breaks SQLAlchemy's `with_loader_criteria` for **all** dialects (PostgreSQL, MySQL, SQLite, etc.), not just Snowflake. The issue occurs when `with_loader_criteria` targets an entity inside a sealed `.subquery()` that contains an ORM `.join()`.

**Minimal reproduction (no database connection needed):**

```python
import sqlalchemy as sa
from sqlalchemy import Column, Integer, String, ForeignKey
from sqlalchemy.orm import DeclarativeBase, with_loader_criteria

class Base(DeclarativeBase):
    pass

class Parent(Base):
    __tablename__ = "parents"
    id = Column(Integer, primary_key=True)
    tenant_id = Column(String, nullable=False)

class Child(Base):
    __tablename__ = "children"
    id = Column(Integer, primary_key=True)
    parent_id = Column(Integer, ForeignKey("parents.id"))
    tenant_id = Column(String, nullable=False)

class GrandChild(Base):
    __tablename__ = "grandchildren"
    id = Column(Integer, primary_key=True)
    child_id = Column(Integer, ForeignKey("children.id"))
    is_active = Column(sa.Boolean, nullable=False, server_default="true")

# --- This single import breaks with_loader_criteria for ALL dialects ---
import snowflake.sqlalchemy  # noqa: F401

# --- Build a query with an ORM join inside a sealed subquery ---
child_subq = (
    sa.select(Child.parent_id, sa.func.count().label("cnt"))
    .join(GrandChild, GrandChild.child_id == Child.id)
    .group_by(Child.parent_id)
    .subquery("child_counts")
)

stmt = (
    sa.select(Parent, child_subq.c.cnt)
    .outerjoin(child_subq, Parent.id == child_subq.c.parent_id)
)

stmt_with_criteria = stmt.options(
    with_loader_criteria(
        GrandChild, GrandChild.is_active == True, include_aliases=True
    )
)

# --- Compile against PostgreSQL (not Snowflake!) ---
from sqlalchemy.dialects import postgresql

compiled = str(stmt_with_criteria.compile(
    dialect=postgresql.dialect(),
    compile_kwargs={"literal_binds": True},
))

has_filter = "is_active" in compiled
print(f"filter injected: {has_filter}")
# Prints: "filter injected: False"
# Expected: "filter injected: True"
```

**Toggle test:** Remove or comment out `import snowflake.sqlalchemy` and the filter is correctly injected.

## 5. What did you expect to see?

`with_loader_criteria` should inject `is_active` into the compiled SQL regardless of whether `snowflake-sqlalchemy` is imported. The Snowflake-specific BCR-1057 handling should only affect queries compiled against the Snowflake dialect.

**What actually happens:** The `is_active` filter is silently dropped from the compiled SQL. No error is raised — the criteria simply aren't applied. This affects any application that has `snowflake-sqlalchemy` installed and imported, even when querying PostgreSQL, MySQL, or SQLite.

### Root cause

`src/snowflake/sqlalchemy/base.py` registers two global `CompileState` plugins:

```python
@CompileState.plugin_for("default", "select")        # replaces ALL core select compilation
class SnowflakeSelectState(SelectState): ...

@sql.base.CompileState.plugin_for("orm", "select")   # replaces ALL ORM select compilation
class SnowflakeORMSelectCompileState(context.ORMSelectCompileState): ...
```

`CompileState.plugin_for` is a **global singleton** — there is no per-dialect dispatch. Once registered, every ORM `select()` in the process compiles through the Snowflake overrides regardless of target dialect.

The overridden `_join_left_to_right` constructs `_Snowflake_ORMJoin` instead of the standard `_ORMJoin`. `_Snowflake_ORMJoin.__init__` (in `util.py`) is a stale copy of `_ORMJoin.__init__` that calls `_Snowflake_Selectable_Join.__init__` directly instead of going through the standard `_ORMJoin.__init__` → `Join.__init__` chain, breaking `_extra_criteria` propagation.

## 6. Debug logs

No relevant log output — the bug is in SQL compilation (silent incorrect behavior), not in connection handling or query execution. There is no error, warning, or log message; the filter is simply absent from the generated SQL.

## Suggested fix

1. The overridden methods in `SnowflakeSelectState` and `SnowflakeORMSelectCompileState` should detect the target dialect and delegate to `super()` for non-Snowflake compilations.

2. `_Snowflake_ORMJoin` and `_Snowflake_Selectable_Join` should delegate to `super().__init__()` instead of reimplementing the parent logic. The BCR-1057 Lateral-without-onclause case can be handled by catching `NoForeignKeysError` from `super()`.

### Validation

Deleting the overridden methods restores correct behavior:

```python
from snowflake.sqlalchemy.base import (
    SnowflakeORMSelectCompileState,
    SnowflakeSelectState,
)

for attr in ("_setup_joins", "_join_determine_implicit_left_side"):
    if attr in vars(SnowflakeSelectState):
        delattr(SnowflakeSelectState, attr)

for attr in ("_join_determine_implicit_left_side", "_join_left_to_right"):
    if attr in vars(SnowflakeORMSelectCompileState):
        delattr(SnowflakeORMSelectCompileState, attr)
```

After this patch, `with_loader_criteria` correctly injects filters on all dialects while the Snowflake dialect still registers and can be used for connections.

gershon-perry:
## Summary

Fixes #676

The BCR-1057 `CompileState.plugin_for` overrides globally replace ORM SELECT compilation for **all** dialects, silently breaking `with_loader_criteria` on non-Snowflake dialects (PostgreSQL, MySQL, SQLite, etc.). The `_extra_criteria` from `with_loader_criteria` is not propagated into entities inside sealed `.subquery()` calls because `_Snowflake_ORMJoin.__init__` is a stale copy of `_ORMJoin.__init__` that bypasses the standard initialization chain.

### Changes

1. **`base.py` — Dialect guards:** Added dialect detection to `SnowflakeSelectState` and `SnowflakeORMSelectCompileState`. All overridden methods now delegate to `super()` for non-Snowflake dialects, ensuring standard SQLAlchemy behavior is fully preserved.

2. **`util.py` — Replace stale copies with thin wrappers:** `_Snowflake_ORMJoin.__init__` and `_Snowflake_Selectable_Join.__init__` now delegate to `super().__init__()` and only intercept `NoForeignKeysError` for the Lateral-without-onclause case (BCR-1057). This eliminates the stale copy problem and ensures future SA improvements (including `_extra_criteria` handling) are inherited automatically.

3. **Regression test:** Added `tests/test_loader_criteria_regression.py` verifying `with_loader_criteria` works correctly on both non-Snowflake and Snowflake dialects.

## Test plan

- [x] Reproduction script confirms `filter injected: True` after fix
- [x] `test_outer_lateral_join` passes (Snowflake lateral join without ON clause still works)
- [x] 100 existing unit tests pass (no regressions)
- [x] New regression test passes for both SQLite and Snowflake dialects
- [ ] Integration tests with live Snowflake account (requires CI)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


gershon-perry:

src/snowflake/sqlalchemy/base.py
```diff
@@ -135,7 +135,15 @@
 # handle Snowflake BCR bcr-1057
 @CompileState.plugin_for("default", "select")
 class SnowflakeSelectState(SelectState):
+    def __init__(self, statement, compiler, **kw):
+        self._is_snowflake = (
+            compiler is not None and compiler.dialect.name == "snowflake"
```
sfc-gh-mraba: There is const `DIALECT_NAME` 
```suggestion
            compiler is not None and compiler.dialect.name == DIALECT_NAME
```

src/snowflake/sqlalchemy/base.py
```diff
@@ -239,24 +246,21 @@ def _join_determine_implicit_left_side(self, raw_columns, left, right, onclause)
 # handle Snowflake BCR bcr-1057
 @sql.base.CompileState.plugin_for("orm", "select")
 class SnowflakeORMSelectCompileState(context.ORMSelectCompileState):
+    _is_snowflake = False
+
+    def _init_global_attributes(self, statement, compiler, **kw):
+        self._is_snowflake = (
+            compiler is not None and compiler.dialect.name == "snowflake"
```
sfc-gh-mraba: There is const `DIALECT_NAME` 
```suggestion
            compiler is not None and compiler.dialect.name == DIALECT_NAME
```

src/snowflake/sqlalchemy/util.py
```diff
@@ -207,138 +209,75 @@ def __init__(
...
+        except NoForeignKeysError:
+            # BCR-1057: Snowflake lateral joins don't require ON clause.
+            # Join.__init__ set self.left and self.right before
+            # _match_primaries raised.
+            if onclause is None and isinstance(self.right, Lateral):
+                self.onclause = None
+                self.isouter = isouter
+                self.full = full
+            else:
+                raise
```
sfc-gh-mraba: Consider different approach — verified against the installed SA source where `expression.Join.__init__` calls `self._match_primaries()` dynamically through the instance:

```python
  class _Snowflake_Selectable_Join(Join):
      def _match_primaries(self, left, right):
          try:
              return super()._match_primaries(left, right)
          except NoForeignKeysError:
              if isinstance(right, Lateral):
                  return None  # BCR-1057: lateral joins don't require FK relationships
              raise
```

This lets `Join.__init__` run to completion cleanly — no manual attribute patching. And because `_ORMJoin.__init__` calls `expression.Join.__init__(self, ...)`, Python dispatches `self._match_primaries()` through the MRO at runtime. So `_Snowflake_ORMJoin` can use multiple inheritance and needs no `__init__` override at all:

```python
  class _Snowflake_ORMJoin(_Snowflake_Selectable_Join, sa_orm_util_ORMJoin):
      pass  # _match_primaries from _Snowflake_Selectable_Join propagates through MRO
```

This replaces ~80 lines of fragile patching with ~10 lines of clean OOP.

sfc-gh-mraba (commented):
Snowflake's dialect needs to support **BCR-1057** — a Snowflake backend change that **forbids an `ON` clause on lateral joins**. Standard SQLAlchemy always requires one. To bridge this, the dialect
   installed three pieces of machinery as global SQLAlchemy plugins:

  1. **`SnowflakeSelectState` / `SnowflakeORMSelectCompileState`** — registered via `CompileState.plugin_for(...)`, which *replaces* SQLAlchemy's compile-state classes for **every** dialect loaded in
   the process (not just Snowflake).
  2. **`_Snowflake_Selectable_Join` / `_Snowflake_ORMJoin`** — subclasses that relaxed the "ON clause required" rule for lateral joins.

  The `_Snowflake_ORMJoin` implementation was a **stale line-by-line copy** of SQLAlchemy's internal `_ORMJoin.__init__`. Over time SQLAlchemy's original had evolved (most importantly, to propagate
  `_extra_criteria` from `with_loader_criteria` into subqueries) while the copy did not.

  ## What Needed To Change

  Two independent problems:

  1. **Cross-dialect leakage (the regression).** Because the plugins are global, any app that merely *imports* snowflake-sqlalchemy silently runs Snowflake's compile logic against
  PostgreSQL/MySQL/SQLite too. Combined with the stale copy above, `with_loader_criteria` filters were dropped inside `.subquery()` constructs on **all** dialects. This is the bug #676 reports.
  2. **Divergence risk.** Because `_Snowflake_ORMJoin` was a stale copy rather than a delegating subclass, every future SQLAlchemy bug fix or feature addition in `_ORMJoin` silently bypasses
  Snowflake users.

  ## Architecture of the Fix

  The fix has two layers:

  **Layer 1 — Dialect gate.** The Snowflake compile-state classes now detect whether the active compilation targets the Snowflake dialect. If not, they short-circuit to `super()` and let standard
  SQLAlchemy run untouched. Non-Snowflake dialects are no longer affected by the import at all.

  **Layer 2 — Thin-wrapper joins.** `_Snowflake_Selectable_Join` and `_Snowflake_ORMJoin` are rewritten from stale copies into minimal subclasses that delegate to SQLAlchemy's originals and intercept
   **only** the lateral-without-ON case. Future SQLAlchemy improvements to `_ORMJoin` now flow through automatically.

  ## Why BCR-1057 Still Matters

  BCR-1057 is a **Snowflake server contract**, not a client choice. The server rejects `JOIN LATERAL ... ON <anything>`. Standard SQLAlchemy unconditionally emits an `ON` clause (matching FK-based
  inference) for any join. Without the lateral-specific intercept, every ORM query using `.lateral()` against Snowflake would fail server-side — so the narrow intercept must remain even after the
  broader cleanup.

  ## Impact on SQLAlchemy 2.0

  Full fix: both the regression (`with_loader_criteria`) and the divergence risk are resolved. Non-Snowflake dialects behave like stock SQLAlchemy. Snowflake behavior is preserved, now inheriting
  future SQLAlchemy fixes automatically.

  ## Impact on SQLAlchemy 1.4

  The dialect gate's hook (`_init_global_attributes`) **does not exist** in SA 1.4, so the gate cannot activate there. SA 1.4 retains the pre-PR behavior — the Snowflake compile path runs
  unconditionally. This is acceptable because:
  - The reported regression (`with_loader_criteria` + `DeclarativeBase`) is an SA 2.0-only API surface — the class doesn't exist in 1.4, so the bug doesn't either.
  - The thin-wrapper refactor of the join classes still applies to 1.4, so SA 1.4 users also benefit from the divergence-risk cleanup.

  The regression test is marked SA 2.0-only for the same reason. No behavior change for existing SA 1.4 users.

sfc-gh-jcieslak (approved):

PR #678, opened 2026-04-20
SNOW-689531: optimize sigle table reflection

sfc-gh-mraba:

   Fixes #SNO689531

Problem

  `get_pk_constraint`, `get_unique_constraints`, `get_foreign_keys`, and `get_indexes` always issued `SHOW … IN SCHEMA` even when called for a single table (e.g. via Inspector or `pandas.read_sql_table()`). On schemas with thousands of tables this causes ~20-second delays per call.

  PR #656 introduced per-table `SHOW … IN TABLE` code paths but gated them behind `cache_column_metadata=True` on
  all SQLAlchemy versions. On SQLAlchemy 2.x this opt-in is unnecessary: `MetaData.reflect()` routes through
  `get_multi_pk_constraint` / `get_multi_unique_constraints` / `get_multi_foreign_keys` / `get_multi_indexes`, so any
  call reaching the singular methods is inherently single-table.

  Fix

  In `_is_single_table_reflection`, move the `IS_VERSION_20` check before the `_cache_column_metadata` guard:

  - SA 2.x: always returns True — the `get_multi_*` hooks own bulk reflection; singular calls are always single-table.
  - SA 1.4: unchanged — `cache_column_metadata=True` opt-in is still required, and the info-cache heuristic still detects `MetaData.reflect()` in progress.

  `get_columns` was already automatic on SA 2.x via the `IS_VERSION_20` guard introduced in #656; this PR applies
  the same logic to the four remaining reflection methods.

### DIFF
--- a/tests/test_index_reflection.py
+++ b/tests/test_index_reflection.py
@@ -6,17 +6,18 @@ from sqlalchemy import MetaData, inspect
 from sqlalchemy.sql.ddl import CreateSchema, DropSchema
 
 from tests.conftest import poll_until
+from tests.util import random_string
 
 
 @pytest.mark.aws
 def test_indexes_reflection(engine_testaccount, db_parameters, sql_compiler):
-    table_name = "test_hybrid_table_2"
-    index_name = "INDEX_NAME_2"
+    table_name = "test_hybrid_table_" + random_string(6)
+    index_name = "INDEX_NAME_" + random_string(6).upper()
     schema = db_parameters["schema"]
     index_columns = ["name", "name2"]
 
     create_table_sql = f"""
-   CREATE OR REPLACE HYBRID TABLE {schema}.{table_name} (
+   CREATE HYBRID TABLE {schema}.{table_name} (
         id INT primary key,
         name VARCHAR,
         name2 VARCHAR,
@@ -33,8 +34,8 @@ def test_indexes_reflection(engine_testaccount, db_parameters, sql_compiler):
         # to propagate before SHOW INDEXES IN TABLE reflects the new index.
         indexes = poll_until(
             lambda: inspect(engine_testaccount).get_indexes(table_name, schema),
-            timeout=15,
-            interval=1,
+            timeout=10,
+            interval=0.5,
         )
 
         assert len(indexes) == 1
--- a/tests/test_index_reflection.py
+++ b/tests/test_index_reflection.py
@@ -5,18 +5,18 @@ import pytest
 from sqlalchemy import MetaData, inspect
 from sqlalchemy.sql.ddl import CreateSchema, DropSchema
 
+from tests.conftest import poll_until
+
 
 @pytest.mark.aws
 def test_indexes_reflection(engine_testaccount, db_parameters, sql_compiler):
-    metadata = MetaData()
-
     table_name = "test_hybrid_table_2"
     index_name = "INDEX_NAME_2"
     schema = db_parameters["schema"]
     index_columns = ["name", "name2"]
 
     create_table_sql = f"""
-   CREATE HYBRID TABLE {table_name} (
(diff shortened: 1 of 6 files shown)


tests/test_index_reflection.py
```diff
@@ -5,18 +5,18 @@
...
-    metadata = MetaData()
-
     table_name = "test_hybrid_table_2"
     index_name = "INDEX_NAME_2"
     schema = db_parameters["schema"]
     index_columns = ["name", "name2"]
 
     create_table_sql = f"""
-   CREATE HYBRID TABLE {table_name} (
+   CREATE OR REPLACE HYBRID TABLE {schema}.{table_name} (
```
sfc-gh-jcieslak: Do we need or replace? I think in this case we need a proper util for identifier generation that wouldn't cause any clashes between test instances, right?

tests/test_index_reflection.py
```diff
@@ -26,19 +26,25 @@ def test_indexes_reflection(engine_testaccount, db_parameters, sql_compiler):
...
-            # Prefixes reflection not supported, example: "HYBRID, DYNAMIC"
-            indexes = insp.get_indexes(table_name, schema)
-            assert len(indexes) == 1
-            assert indexes[0].get("name") == index_name
-            assert indexes[0].get("column_names") == index_columns
+        # On AWS, Hybrid Table (Unistore) index metadata can take a few seconds
+        # to propagate before SHOW INDEXES IN TABLE reflects the new index.
+        indexes = poll_until(
+            lambda: inspect(engine_testaccount).get_indexes(table_name, schema),
+            timeout=15,
```
sfc-gh-jcieslak: let's maybe change it to less time out and more intervals? Because now, due to the propagation the test is guaranteed to run for at least 15 seconds (and could run faster with different timeout settings)

tests/test_unit_core.py
```diff
@@ -217,3 +228,162 @@ def test_get_server_version_info_parsing(raw_value, expected):
...
+        dialect = _make_dialect(_cache_column_metadata=True)
+        assert dialect._is_single_table_reflection("PUBLIC", info_cache={}) is True
+
+
+# ---------------------------------------------------------------------------
+# Single-table dispatch on SA 2.x (SNOW-689531)
+# ---------------------------------------------------------------------------
+
+
+class TestSingleTableDispatchSA2:
```
sfc-gh-jcieslak, line 288: Most added tests could be grouped into parametrized tests as I can see repetitions across tests

sfc-gh-mraba: Collapsed the four `test_get_*_uses_table_path` tests into a single `test_uses_table_path`. 
The `test_get_indexes_passes_raw_tablename_not_normalized` test is left separate since it covers a distinct concern (type of the table_name argument, not dispatch routing).

tests/test_index_reflection.py
sfc-gh-mraba: Changed to `timeout=10, interval=0.5` — twice as frequent, slightly tighter bound. 
Note: `poll_until` already handles the falsy-`[]` case by returning the last result once the deadline passes, so a missing index still surfaces (just up to 10 s later rather than immediately); this is acceptable given the async propagation nature of the test.

tests/test_index_reflection.py
sfc-gh-mraba: Switched to `random_string(6)`-suffixed names (consistent with `test_decfloat.py` / `test_timestamp.py`) and dropped `OR REPLACE`. A leftover table from a crashed run will now cause an explicit `CREATE` failure rather than silently overwriting the fixture state.

sfc-gh-jcieslak: 👍, but this is a general project issue imo, and such a generator should be introduced to avoid such situations. Something for later.

sfc-gh-jcieslak (approved):

PR #690, opened 2026-05-25
SNOW-3570860 move to sa 20

sfc-gh-mraba:

   Fixes #SNOW-3570860

- Drop SQLAlchemy 1.4 compatibility code and require SQLAlchemy 2.0+.
- Update supported Python versions to 3.9 through 3.14.
- Remove obsolete SQLAlchemy 1.4 test paths, CI matrix entries, and compatibility helpers.
- Keep UUID reflection available on SQLAlchemy 2.0 and return UUID values as strings.
- Clean up SQLAlchemy 2-only reflection and ORM compiler paths.
- Move unreleased changelog entries under the v2.0.0 section.


.github/workflows/build_test.yml
```diff
@@ -397,7 +255,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: ${{ matrix.python-version }}
+          python-version: "3.13"
```
sfc-gh-jcieslak: Is it intended to use only one Python version in tests?

src/snowflake/sqlalchemy/snowdialect.py
```diff
@@ -172,7 +170,6 @@ def __init__(
         isolation_level: Optional[str] = SnowflakeIsolationLevel.READ_COMMITTED.value,
         enable_decfloat: bool = False,
         case_sensitive_identifiers: bool = False,
-        cache_column_metadata: bool = False,
```
sfc-gh-jcieslak, line 216: Q (for potential future PRs): Do we have other flags that could be removed in favor of better defaults?

src/snowflake/sqlalchemy/snowdialect.py
```diff
@@ -1710,20 +1635,14 @@ def _get_table_indexes(self, connection, table_name, schema, **kw):
...
-        if self._is_single_table_reflection(schema, **kw):
-            # Pass the raw string so _always_quote_join can correctly
-            # denormalize (uppercase) it.  normalize_name() wraps plain
-            # lowercase strings in quoted_name(quote=True), and
-            # quoted_name.upper() is intentionally a no-op for quote=True
-            # objects, which would cause _always_quote_join to emit
-            # "table_name" (case-sensitive lowercase) instead of the
-            # correct "TABLE_NAME" (case-insensitive uppercase).
-            return self._get_table_indexes(connection, str(tablename), schema, **kw)
-        table_name = self.normalize_name(str(tablename))
```
sfc-gh-jcieslak, line 1890: Q: Do we have all `..multi..` hooks handled properly? E.g., for indexes it seems a bit different from here, which is why I'm wondering if those methods should also be adjusted.

tests/sqlalchemy_test_suite/test_suite.py
```diff
@@ -71,6 +77,24 @@ def test_no_results_for_non_returning_insert(self, connection, style, executeman
         pass
 
 
+class TrueDivTest(_TrueDivTest):
+    @pytest.mark.skip("`//` not supported")
+    def test_floordiv_integer_bound(self, connection):
```
sfc-gh-jcieslak, line 82: Q: Just so I understand correctly, we're skipping those as we have our own tests that work with the dialect-level flag, so we don't need the tests provided by SQA for truediv, correct?

sfc-gh-jcieslak (approved):

.github/workflows/build_test.yml
sfc-gh-mraba: I think it is. What I think I will do is to switch to lowest supported.

src/snowflake/sqlalchemy/snowdialect.py
sfc-gh-mraba: This is open for discussions before RC release.

src/snowflake/sqlalchemy/snowdialect.py
sfc-gh-mraba: Yes, all multi hooks follow the same two-path pattern:

- `get_<thing>` → per-table `SHOW … IN TABLE` (called by Inspector for single-table inspection, e.g. `inspector.get_indexes()` or `pandas.read_sql_table()`)
- `get_multi_<thing>` → schema-wide `SHOW … IN SCHEMA` (called by SA 2.x for `MetaData.reflect()`)

Indexes, PK, UK, and FK are all consistent. The apparent difference from main is that `get_indexes` no longer has the `_is_single_table_reflection()` branch — that helper was removed as part of the 1.4 drop, since on SA 2.x the framework itself routes bulk reflection to `get_multi_*`, so the branching in the singular methods was never needed.

tests/sqlalchemy_test_suite/test_suite.py
sfc-gh-mraba: Good catch — the skip reason is incorrect. SA compiles `//` to `FLOOR(x / y)`, not a SQL `//` operator, so "Snowflake does not provide `//`" is wrong. Snowflake supports `FLOOR()` fine.

The real issue is `force_div_is_floordiv=True` (the current default):
- truediv `/` tests expect `15/10 = 1.5` — these pass because `visit_truediv_binary` delegates to SA's base which emits `15 / CAST(10 AS NUMERIC) = 1.5`
- floordiv `//` tests expect `15//10 = 1` — these would also pass via `FLOOR(15 / 10) = 1`, just with a `PendingDeprecationWarning`

So the skips are likely unnecessary. However, this is tied to the open `force_div_is_floordiv` discussion — if we flip its default to `False` in v2.0.0 (which I'm leaning towards given we're doing a major version bump), the warning disappears and the entire `TrueDivTest` subclass can be dropped. Leaving this pending until Q2 is resolved.

PR #693, opened 2026-05-28
SNOW-3480955: table reflection improvements

sfc-gh-mraba:

   Fixes #SNOW-3480955

Fix inconsistent identifier quoting in the `DESC TABLE` structured-type fallback path
- `_StructuredTypeInfoManager.get_table_columns` was the only reflection path that built a `DESC TABLE` identifier string via raw f-string concatenation on denormalised names. Every other path in the dialect applies `ip.quote(denormalize_name(...))` — either directly or through `_always_quote_join`. Schema and table components are now double-quoted consistently before the statement is constructed, matching the established pattern.
- `get_table_columns` is simplified by delegating to the existing `get_table_columns_by_full_name` once quoting is applied. This collapses two previously divergent code paths (the SA 2.x / `cache_column_metadata=True` path and the SA 1.4 fallback) into one.
- Two tests in `test_structured_datatypes.py` (`test_structured_type_on_dropped_table`, `test_structured_type_on_table_with_map`) were constructed with `_NameUtils(None)` and a `denormalize_name`
  patch to compensate. The patch masked a latent dependency on a real `identifier_preparer` that the quoting alignment exposed. Both tests now use `_NameUtils(SnowflakeDialect().identifier_preparer) `and drop the unnecessary patch, making them consistent with the rest of the test suite.
- Added `tests/test_unit_structured_type_info_manager.py` with 10 unit tests covering quoting correctness for plain, uppercase, mixed-case, dotted, and special-character identifiers in both the schema and table name positions.

### DIFF
--- a/tests/test_structured_datatypes.py
+++ b/tests/test_structured_datatypes.py
@@ -25,6 +25,7 @@ from snowflake.sqlalchemy import NUMBER, IcebergTable, SnowflakeTable
 from snowflake.sqlalchemy.custom_types import ARRAY, MAP, OBJECT, TEXT
 from snowflake.sqlalchemy.exc import StructuredTypeNotSupportedInTableColumnsError
 from snowflake.sqlalchemy.name_utils import _NameUtils
+from snowflake.sqlalchemy.snowdialect import SnowflakeDialect
 from snowflake.sqlalchemy.structured_type_info_manager import _StructuredTypeInfoManager
 
 
@@ -603,14 +604,10 @@ def test_structured_type_not_supported_in_table_columns_error(
 
 
 @patch.object(_StructuredTypeInfoManager, "_execute_desc")
-@patch.object(_NameUtils, "denormalize_name")
-def test_structured_type_on_dropped_table(
-    mocked_denormalize_name_method, mocked_execute_desc_method
-):
+def test_structured_type_on_dropped_table(mocked_execute_desc_method):
     mocked_execute_desc_method.return_value = None
-    mocked_denormalize_name_method.side_effect = lambda v: v
     structured_type_info = _StructuredTypeInfoManager(
-        None, _NameUtils(None), "mySchema"
+        None, _NameUtils(SnowflakeDialect().identifier_preparer), "mySchema"
     )
     result = structured_type_info.get_column_info(
         "mySchema", "dropped_table", "structured_type_col"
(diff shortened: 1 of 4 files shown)


tests/test_structured_datatypes.py
```diff
@@ -603,14 +604,10 @@ def test_structured_type_not_supported_in_table_columns_error(
...
-@patch.object(_NameUtils, "denormalize_name")
-def test_structured_type_on_dropped_table(
-    mocked_denormalize_name_method, mocked_execute_desc_method
-):
+def test_structured_type_on_dropped_table(mocked_execute_desc_method):
     mocked_execute_desc_method.return_value = None
-    mocked_denormalize_name_method.side_effect = lambda v: v
     structured_type_info = _StructuredTypeInfoManager(
-        None, _NameUtils(None), "mySchema"
+        None, _NameUtils(SnowflakeDialect().identifier_preparer), "mySchema"
```
sfc-gh-jcieslak, line 610: Nit: do we expect identifier preparers other than the Snowflake one from SnowflakeDialect here? If not, _NameUtils should probably default to the Snowflake identifier preparer. Alternatively, if _NameUtils should stay generic, _StructuredTypeInfoManager could instantiate _NameUtils implicitly with the Snowflake identifier preparer instead of expecting callers to pass a variant we don't actually support.

tests/test_unit_structured_type_info_manager.py
```diff
@@ -0,0 +1,204 @@
...
+    payload = "myschema; SELECT 1--"
+    manager, captured = _make_manager()
+    manager.get_table_columns("mytable", schema=payload)
+
+    sql = _first_sql(captured)
+    assert (
+        f'"{payload}"' in sql
+    ), f"Schema payload not enclosed in double-quotes; got: {sql!r}"
+    stripped = re.sub(r'"[^"]*"', "", sql)
+    assert (
```
sfc-gh-jcieslak, line 187: nit: I don't this assert is needed (as in other tests) because above says all we wanted to know

sfc-gh-jcieslak (approved):

tests/test_structured_datatypes.py
sfc-gh-mraba: Good point. In practice `_StructuredTypeInfoManager` is always constructed with `self.name_utils` from `SnowflakeDialect`, so there is only ever one valid input.

The reason the parameter cannot be eliminated or defaulted internally is that `_NameUtils` carries the per-engine `case_sensitive_identifiers` flag (added in SNOW-1232488/#675). If `_StructuredTypeInfoManager` created its own `_NameUtils` internally — e.g. `_NameUtils(SnowflakeDialect().identifier_preparer)` — it would always use `case_sensitive_identifiers=False`, silently ignoring the user's engine-level setting and causing structured-type column names to reflect as plain str instead of `quoted_name(..., True)` for users who opted in.

So the parameter needs to stay. I'll add a docstring note on `_StructuredTypeInfoManager` clarifying that callers must always pass `self.name_utils` (i.e. the dialect's own instance) rather than constructing a fresh one.

tests/test_unit_structured_type_info_manager.py
sfc-gh-mraba: Good catch. Worth removing.

PR #743, opened 2026-07-29
SNOW-1023317: Support RELY on PRIMARY KEY / FOREIGN KEY / UNIQUE constraints

### LINKED ISSUE #463: Add support for RELY
## Description : 
Currently we can't define on Columns the option : snowflake_rely=True unless we
use @compiles decorator on CreateTable or we use event.listen after tables creation

## What is the desired behavior?
The ability to set rely on constraints when defining a table like this : 
Table('table_name', metadata,
             Column('col_name', col_type, primary_key=True, snowflake_rely=True),
             Column('col_name', col_type, ForeignKey("t.col"), snowflake_rely=True)
             )

## How would this improve `snowflake-connector-python`?
Add the functionality

sfc-gh-mraba:

   Fixes #463

   Snowflake does not enforce PRIMARY KEY / UNIQUE / FOREIGN KEY constraints; the `RELY` property lets the optimizer trust them for query rewrites (e.g. join elimination). This adds an opt-in `snowflake_rely=True` dialect keyword on those constraints. `SnowflakeDDLCompiler` overrides `visit_primary_key_constraint`, `visit_foreign_key_constraint`, and `visit_unique_constraint` to call `super()` and append ` RELY` when `constraint.dialect_options["snowflake"].get("rely")` is truthy (mirroring the existing `snowflake_clusterby` pattern). It defaults off, so generated DDL and query results are unchanged unless the user explicitly opts in — important because `RELY` is an unverified promise and, if the data violates the constraint, the optimizer could return incorrect results. Added unit tests in `tests/test_rely.py` (PK/FK/UNIQUE positive, default-off, and `rely=False`) and README documentation with a correctness warning.

---

> [!NOTE]
> **Medium Risk**
> Opt-in DDL only, but enabling RELY on constraints that data does not actually satisfy can cause the Snowflake optimizer to return incorrect results; the change is limited to constraint compilation, not runtime queries.
> 
> **Overview**
> Adds **opt-in `snowflake_rely=True`** on `PrimaryKeyConstraint`, `ForeignKeyConstraint`, and `UniqueConstraint` so compiled DDL appends Snowflake **`RELY`** (e.g. `PRIMARY KEY (id) RELY`), enabling the optimizer to trust metadata-only constraints for rewrites such as join elimination.
> 
> `SnowflakeDDLCompiler` delegates to the base constraint visitors and appends ` RELY` when the Snowflake dialect option `rely` is set; it stays **off by default**, so existing DDL is unchanged unless callers opt in.
> 
> README and release notes document usage and warn that **`RELY` is unverified**—bad data can yield **wrong query results** if the optimizer relies on a false assumption.
> 
> Adds **`tests/test_rely.py`** covering PK/FK/UNIQUE with `snowflake_rely=True`, default-off behavior, and `snowflake_rely=False` (fixes #463 / SNOW-1023317).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 27b3d81748daab19d4519b3ae2ab69bbbc200174. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/DESCRIPTION.md
+++ b/DESCRIPTION.md
@@ -9,6 +9,7 @@ Source code is also available at:
 
 # Unreleased Notes
 
+- Add opt-in `snowflake_rely=True` to render `RELY` on `PRIMARY KEY` / `FOREIGN KEY` / `UNIQUE` constraints so the optimizer can trust them for query rewrites such as join elimination (SNOW-1023317 / GH #463).
 - Document SSO/Okta authentication via the `authenticator` connect_args parameter (SNOW-715550).
 - Document how to enable connector bulk array binding for large `executemany` inserts via `qmark` paramstyle (SNOW-710474).
 - Document using a SQLAlchemy `VALUES` source with `MergeInto` (no staging table needed) (SNOW-889678 / GH #435).
--- a/README.md
+++ b/README.md
@@ -946,6 +946,37 @@ stmt = select(my_table).where(
 )
 ```
 
+### RELY Constraint Property
+
+Snowflake does not enforce `PRIMARY KEY`, `UNIQUE`, or `FOREIGN KEY` constraints — they are metadata only. The `RELY` property tells the query optimizer it may **trust** such a constraint for rewrites like join elimination, even though it isn't enforced. Set the opt-in `snowflake_rely=True` dialect keyword on a constraint to emit it:
+
+```python
+from sqlalchemy import (
+    Column, ForeignKeyConstraint, Integer, MetaData,
+    PrimaryKeyConstraint, String, Table, UniqueConstraint,
+)
+
+metadata = MetaData()
+
+dim = Table(
+    "dim", metadata,
+    Column("id", Integer),
+    Column("email", String),
+    PrimaryKeyConstraint("id", snowflake_rely=True),
+    UniqueConstraint("email", snowflake_rely=True),
+)
+# ... PRIMARY KEY (id) RELY ... UNIQUE (email) RELY
+
+fact = Table(
+    "fact", metadata,
+    Column("dim_id", Integer),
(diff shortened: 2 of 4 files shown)


sfc-gh-mraba:
Thanks, but this is a false positive for this dialect. `SnowflakeDialect.construct_arguments` is `None` (inherited from `DefaultDialect`), which means SQLAlchemy accepts arbitrary `snowflake_*` keyword arguments without validation rather than dropping them. Verified:

```python
>>> SnowflakeDialect.construct_arguments is None
True
>>> pkc = PrimaryKeyConstraint("id", snowflake_rely=True)   # attached to a Table
>>> pkc.dialect_options["snowflake"]["rely"]
True
```

This is the exact same mechanism the existing `snowflake_clusterby` option relies on (the module-level `construct_arguments` list in `base.py` is not wired to the dialect). The added tests in `tests/test_rely.py` compile real `CreateTable` DDL end-to-end and assert `PRIMARY KEY (id) RELY`, `REFERENCES parent (id) RELY`, and `UNIQUE (email) RELY` are emitted, so the kwarg demonstrably reaches the compiler.

Registering via `DialectKWArgs.argument_for(...)` isn't viable here: it raises `ArgumentError` when the dialect's `construct_arguments` is `None`, and defining `construct_arguments` to enable per-arg validation would flip the dialect into restrictive mode and reject the existing arbitrary `snowflake_*` kwargs (`clusterby`, custom table options). The code reads the value defensively via `.get("rely")`, so an unset option is simply `None` (no `RELY`).

sfc-gh-jcieslak (approved):

PR #750, opened 2026-08-03
SNOW-884382: Add row_access_policy table option to SnowflakeTable

### LINKED ISSUE #432: Add support for setting row_access_policy
## What is the current behavior?

The SnowDialect does not support defining a row access policy.

## What is the desired behavior?

To be able to define a row access policy using sqlalchemy.Table

## References, Other Background

This is similar to specifying `snowflake_clusterby`.

sfc-gh-mraba:

   Fixes #432

Snowflake row access policies could not be declared through the dialect. This PR adds a `RowAccessPolicyOption` to the existing table-option framework and wires it into `SnowflakeTable`, so users can write`SnowflakeTable(..., row_access_policy=("my_policy", ["col1", "col2"]))` (or pass an explicit `RowAccessPolicyOption`) to render `WITH ROW ACCESS POLICY my_policy ON (col1, col2)` in the generated `CREATE TABLE`.

The option follows the same pattern as the other table options (e.g. cluster by), validates its input type, and is purely additive/opt-in — tables without the option are unaffected. Compile/snapshot tests in
 `tests/custom_tables/test_compile_snowflake_table.py` cover the tuple and explicit forms, combination with cluster by, and invalid-type validation.

---

> [!NOTE]
> **Low Risk**
> Additive DDL-only feature with no change to existing tables; main risk is incorrect policy/column names in generated SQL, caught by tests and user review of migrations.
> 
> **Overview**
> Adds **row access policy** support on `SnowflakeTable` so `CREATE TABLE` can emit `WITH ROW ACCESS POLICY … ON (…)`.
> 
> Users pass `row_access_policy` as a `(policy_name, [columns])` tuple or a `RowAccessPolicyOption`. Policy names may be simple, fully qualified (`db.schema.policy`), or an explicit `FQN`; rendering uses per-part quoting aligned with the dialect preparer. The clause is ordered after `CLUSTER BY` via table-option priority.
> 
> A new `FQN` helper parses `[[database.]schema.]name` using the same `split_identifier_parts` scanner as the preparer. `RowAccessPolicyTableBase` wires the option into the custom table hierarchy (`TableFromQueryBase` → `SnowflakeTable`). Public exports include `FQN` and `RowAccessPolicyOption`; README and release notes document usage. Compile snapshots and unit tests cover DDL output, validation, and FQN parsing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cef83c6e876250baa1435fe857fcd2a66ea86255. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/snowflake/sqlalchemy/sql/custom_schema/options/row_access_policy_option.py
```diff
@@ -0,0 +1,97 @@
...
+    """
+
+    def __init__(self, policy_name: str, columns: list[str]) -> None:
+        super().__init__()
+        self.policy_name: str = policy_name
+        self.columns: list[str] = columns
+
+    @property
+    def priority(self) -> Priority:
+        return Priority.HIGH
```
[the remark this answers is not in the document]
sfc-gh-mraba: Good catch — fixed in 9087ae7. `RowAccessPolicyOption.priority` was `HIGH` (same as `ClusterByOption`), and the option_name tiebreaker rendered `WITH ROW ACCESS POLICY` before `CLUSTER BY`. Lowered it to `Priority.LOW` so `CLUSTER BY` is emitted first, matching Snowflake's CREATE TABLE grammar. Combined DDL is now `... CLUSTER BY (id) WITH ROW ACCESS POLICY my_policy ON (geom, name)`, and the snapshot test was updated accordingly.

src/snowflake/sqlalchemy/sql/custom_schema/options/row_access_policy_option.py
```diff
@@ -0,0 +1,100 @@
...
+
+class RowAccessPolicyOption(TableOption):
+    """Class to represent a row access policy option in Snowflake Tables.
+
+    Takes a tuple of (policy_name, list_of_columns) where:
+    - policy_name: str - name of the row access policy
+    - list_of_columns: list[str] - list of columns the policy applies to
+
+    Example:
+        row_access_policy = RowAccessPolicyOption('my_policy', ['col1', 'col2'])
```
sfc-gh-jcieslak, line 27: Question: shouldn't policy_name be a FQN? or does row access policy has to be in the same schema as table to be applied? Anyway, I think we could require FQN to be passed in new schema-level fields as it may simplify other things in the future (as we don't assume a good database and schema is used in the driver context)

sfc-gh-mraba: Good question. A row access policy does **not** have to live in the table's schema — Snowflake resolves an unqualified `policy_name` against the session's current database/schema, but you can reference a policy in another database/schema by qualifying it. Since the dialect doesn't assume a "good" database/schema in the driver context, supporting a qualified reference matters.

Addressed in 48ce6cd: `policy_name` now accepts `str | FQN`.
- A string is parsed as `[[database.]schema.]name` and each identifier part is quoted independently (already-quoted parts like `"My Policy"` are preserved), so `"gov_db.security.pii_policy"` renders as `gov_db.security.pii_policy`.
- You can also pass an explicit `FQN` (now exported as `snowflake.sqlalchemy.FQN`): `RowAccessPolicyOption(FQN.from_string("gov_db.security.pii_policy"), [...])`.

I kept it `str | FQN` rather than FQN-only so the existing `("my_policy", [cols])` tuple API stays backward compatible, while still allowing full qualification via either a dotted string or an `FQN`. The `FQN` helper is standalone (adapted from the CLI's identifier parsing) so it can back other schema-level fields later, per your suggestion. Covered by new unit tests for FQN parsing and RAP rendering (qualified string, quoted part, and FQN object).

sfc-gh-jcieslak: The thing is, we don't need to stay backward compatible, as it's a new field. I'm just asking / proposing if new fields should be FQN-only to stop guessing databases and schemas (we have now different levels we could be using in the dialect) and start enforcing full names to cut ourselves from those levels of guessing. I'm good with proceeding with this approach, but we could start leaning toward the FQN approach with those new fields.

sfc-gh-jcieslak (approved):

src/snowflake/sqlalchemy/sql/custom_schema/options/row_access_policy_option.py
sfc-gh-mraba: Agreed on the FQN direction — I dug into where it should apply and landed on a concrete recommendation:

1. **Consistency is the deciding factor.** Making this field FQN-only would make `row_access_policy` the *only* identifier field that rejects a plain string, while every sibling accepts one: `warehouse` (`IdentifierOption`), Iceberg `external_volume`/`catalog` (`LiteralOption`), and the stage/`format_name` identifiers in the custom commands. So FQN-only here creates an inconsistent island rather than a convention. I'd rather keep `str | FQN` here and, in a follow-up, offer `str | FQN` across the identifier-bearing fields — that gives the typed/validated FQN option everywhere without forcing `FQN.from_string("x")` for trivial names. FQN-only makes the most sense reserved for brand-new fields (masking policies, tags, projection/aggregation policies) where there's no established string API to stay consistent with — that's where your "stop guessing db/schema" argument lands cleanest.

2. **The prerequisite is done.** FQN now tokenizes via the dialect's own `split_identifier_parts` scanner and stores each part as `quoted_name(value, quote=...)`, matching the preparer's model. This fixed a latent double-quoting bug (an already-quoted part like a quoted `Sec` would have become triple-quoted if FQN parts were fed through the shared quoting path), so FQN is now safe to reuse beyond its own render.

3. **It's render-policy-agnostic**, so adoption elsewhere can be byte-for-byte SQL-preserving: FQN just holds the parts, and each call site keeps its current renderer (RAP → `preparer.quote`; stages/formats → `quote_identifier_if_unsafe`). Best first targets are `ExternalStage`/`CreateStage` namespace+name and `CreateFileFormat.format_name` (retires the ad-hoc `prepare_namespace` munging, unifying on the one scanner).

Proposal: keep #750 as `str | FQN`, and scope the broader adoption + the FQN-only-for-new-fields policy as a separate PR.

PR #755, opened 2026-08-07
Fix index related bugs

### LINKED ISSUE #754: [Bug / Discussion] 3 bugs relating to indices
I've filled out the bug form below, but I think it is better to explain the three bugs I've encountered.

I encountered problems when trying to `--autogenerate` Alembic migrations that use hybrid tables. I wasn't sure if this was an Alembic quirk or what at first. But I dug deeper and realized it was a series of multiple issues related to index handling in snowflake-sqlalchemy.

## 1. Case sensitivity for INDEX identifiers

Index names were not being normalized, and thus a lowercase index in my code would be dropped and then rebuilt. E.g. I would see something like this in the autogeneration:

```python
def upgrade() -> None:
    op.drop_index(op.f("ix__foo__bar_id"), table_name="foo")
    op.create_index(op.f("ix__foo__bar_id"), table_name="foo")
```

I realized the reason for this is because index names are not being normalized, unlike basically all other identifiers for objects in Snowflake (table names, column names, uniqueness constraints, schemas, and so on).

It seems _everywhere else_ in `snowflake-sqlalchemy` has opted into a sort of "contract" that identifiers are case-insensitive (or more specifically are "normalized") _except_ for indices.

And this has the consequence of breaking the round-trip from Python -> Snowflake -> Python because the returned value `IX__FOO__BAR_ID` is no longer equal to `ix__foo__bar_id`.

For that reason, to me, this seems like a bug.

## 2. `Table(..., autoload_with=engine)` does not tie indices to tables when using the default schema

TLDR of this bug - `Table(..., autoload_with=engine)` does not tie indices to tables when no schema is explicitly defined (i.e. the connection level default schema is used) because `get_multi_indexes` return `('public', name)` instead of `(None, name)`. In this case, I am using `public` as my default schema. The issue though is that these tuples need to key on _what is specific in the SqlAlchemy model_, and since my input at the model level is `None`, the metadata reflection step during Alembic automigrations miss this as well.

## 3. `_parse_index_rows()` returns `['']` instead of `[]` when there are no `INCLUDE` columns specified

This one is relatively straightforwardly identifiable as a bug and easy to show with a minimal reproduction (AI generated example here):

```python
import sqlalchemy as sa
from snowflake.sqlalchemy import HybridTable

engine = sa.create_engine(...)  # any account/database/schema

metadata = sa.MetaData()
HybridTable(
    "include_columns_repro",
    metadata,
    sa.Column("id", sa.Integer, primary_key=True),
    sa.Column("val", sa.Integer),
    sa.Index("ix_include_columns_repro_val", "val"),  # no INCLUDE columns
)

metadata.create_all(engine)
try:
    (index,) = sa.inspect(engine).get_indexes("include_columns_repro")
    print("include_columns:", index["include_columns"])  # [''] — expected []
finally:
    metadata.drop_all(engine)
```

Output:

```
include_columns: ['']
```

---
 
## The form (for issue 1 specifically)

   3.14.3

   macOS-26.5.2-arm64-arm-64bit-Mach-O

   snowflake-connector-python==4.6.0
   snowflake-sqlalchemy==1.11.0
   sqlalchemy==2.0.51

4. What did you do?

   **(AI generated minimally reproducible example)**

   ```python
   import sqlalchemy as sa
   from snowflake.sqlalchemy import HybridTable
   
   engine = sa.create_engine(...)  # any account/database/schema
   
   metadata = sa.MetaData()
   HybridTable(
       "index_name_repro",
       metadata,
       sa.Column("id", sa.Integer, primary_key=True),
       sa.Column("val", sa.Integer),
       sa.Index("ix_index_name_repro_val", "val"),
       sa.UniqueConstraint("val", name="uq_index_name_repro_val"),
   )
   metadata.create_all(engine)
   try:
       insp = sa.inspect(engine)
       print("columns:", [c["name"] for c in insp.get_columns("index_name_repro")])
       print("uniques:", [u["name"] for u in insp.get_unique_constraints("index_name_repro")])
       print("indexes:", [i["name"] for i in insp.get_indexes("index_name_repro")])
   finally:
       metadata.drop_all(engine)
   ```

   Output:
   ```
   columns: ['id', 'val']
   uniques: ['uq_index_name_repro_val']
   indexes: ['UQ_INDEX_NAME_REPRO_VAL', 'IX_INDEX_NAME_REPRO_VAL']
   ```
 
   I expected to see lowercase identifiers for the indices.

6. Can you set logging to DEBUG and collect the logs?

   I think the above reproducion should be enough 🙂 

---

## One more issue?

There is no way to await index creation in Snowflake from an index DDL statement. This causes problems when generating multiple indices. We have solved this via our `SnowflakeImpl`. However, I do not see a clean way to fix this in snowflake-sqlalchemy-- this is purely an Alembic side operation, and Alembic plugins don't allow for dialect-specific overrides of operations. The only fix I can see would be to create a custom `SnowflakeImpl` instead of asking users to subclass the `DefaultImpl` in their `env.py`.

dwreeves:

   Fixes #754

   Fixes all 3 related issues via normalizing names, handling `"[]"` correctly, and handling `schema=None` correctly in the return value to `get_multi_indexes()`.

   Note: library code was hand-written, but unit-tests were largely AI generated.

---

> [!NOTE]
> **Medium Risk**
> Changes reflection metadata shape and naming for hybrid indexes, which can affect autoload and migration diffs, but scope is limited to index parsing and bulk reflection keying.
> 
> **Overview**
> Fixes **hybrid table index reflection** so `MetaData.reflect()` and `Inspector.get_indexes()` return consistent, usable metadata (closes #754).
> 
> **`parse_index_columns`** now returns an empty list for `"[]"` instead of a spurious empty string entry, which affected **include columns** when Snowflake reports no included columns.
> 
> **`_parse_index_rows`** applies **`normalize_name`** to reflected index names so unquoted identifiers match SQLAlchemy’s lowercase convention; quoted/case-sensitive names stay as returned.
> 
> **`get_multi_indexes`** resolves the default schema internally via `resolved_schema` but keeps the **original `schema` argument** (including `None`) in result keys, aligning with other bulk reflection hooks so indexes attach when reflecting without an explicit schema.
> 
> Integration tests cover uppercase/lowercase index names, default vs explicit schema, and INCLUDE columns; new unit tests cover parsing, normalization, PK sentinel filtering, and multi-index keying.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e3a4c8b8e9f1846643db3415094db0dfb9701bda. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/snowflake/sqlalchemy/parser/custom_type_parser.py
+++ b/src/snowflake/sqlalchemy/parser/custom_type_parser.py
@@ -133,7 +133,10 @@ def parse_index_columns(columns: str) -> list[str]:
     :example:
         For input `"[A, B, C]"`, the output is `['A', 'B', 'C']`.
     """
-    return [column.strip() for column in columns.strip("[]").split(",")]
+    inner = columns.strip("[]")
+    if not inner:
+        return []
+    return [column.strip() for column in inner.split(",")]
 
 
 def parse_type(type_text: str) -> TypeEngine:
--- a/src/snowflake/sqlalchemy/snowdialect.py
+++ b/src/snowflake/sqlalchemy/snowdialect.py
@@ -1932,17 +1932,18 @@ class SnowflakeDialect(default.DefaultDialect):
         """
         Gets the indexes definition
         """
-        schema = schema or self.default_schema_name
+        resolved_schema = schema or self.default_schema_name
         hybrid_table_names = self.get_table_names_with_prefix(
             connection,
-            schema=schema,
+            schema=resolved_schema,
             prefix=CustomTablePrefix.HYBRID.name,
             info_cache=kw.get("info_cache", None),
         )
+
         if len(hybrid_table_names) == 0:
             return []
 
-        full_schema_name = self._get_full_schema_name(connection, schema, **kw)
+        full_schema_name = self._get_full_schema_name(connection, resolved_schema, **kw)
         result = connection.execute(
             text(
                 f"SHOW /* sqlalchemy:get_multi_indexes */ INDEXES IN SCHEMA {full_schema_name}"
@@ -1984,7 +1985,7 @@ class SnowflakeDialect(default.DefaultDialect):
             table_name = self.normalize_name(str(row[name_to_index_map["table"]]))
             indexes[table_name].append(
                 {
-                    "name": row[name_to_index_map["name"]],
+                    "name": self.normalize_name(row[name_to_index_map["name"]]),
                     "unique": row[name_to_index_map["is_unique"]] == "Y",
                     "column_names": [
                         self.normalize_name(column)
--- a/tests/custom_tables/test_reflect_hybrid_table.py
+++ b/tests/custom_tables/test_reflect_hybrid_table.py
@@ -5,6 +5,8 @@ import pytest
 from sqlalchemy import MetaData, Table
 from sqlalchemy.sql.ddl import CreateTable
 
+from tests.util import random_string
+
 
 @pytest.mark.aws
 def test_simple_reflection_hybrid_table_as_table(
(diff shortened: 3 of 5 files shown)


sfc-gh-mraba:
Landed with https://github.com/snowflakedb/snowflake-sqlalchemy/commit/6da9a6c329dff0db37ee438c681e041a5189b200

dwreeves:
Thanks!

